Issue #49, opened 2022-04-15
@daohaus/commons: Set up Sage Library


skuhlmann:
## Overview

Create utilities library in the sage mono repo.

### Todo

- [x] generate package w/ name: @daohaus/utilities
- [x] see spec/rage repo for folder structure

### resources
https://hackmd.io/wtDwRLeVTsepeDjeiQ0lyg?view

jonathanprozzi:
@skuhlmann Moving this into Done! Won't close them out until our Thursday Dragon Sync but moving it over since the PR is merged!

PR #70, opened 2022-04-19
Feature/gen commons pkg

skuhlmann:
## GitHub Issue

[GitHub issue.](https://github.com/HausDAO/daohaus-monorepo/issues/49)

## Changes

- nx gen the commons library
- directory setup: constants, legos, utils, types

## Checks

- [x] App runs locally


libs/commons/src/index.ts
```diff
@@ -0,0 +1,4 @@
+// export * from './lib/constants';
+// export * from './lib/legos';
+// export * from './lib/types';
+// export * from './lib/utils';
```
alexkeating, line 4: This should be removed, probably along with the empty files

alexkeating (commented):
Looks good, once the commented out code and empty files are removed this should be good to merge

libs/commons/src/index.ts
skuhlmann: You mind if I add one util/example file for each? I've got to bring them back right away.

skuhlmann: Actually I just removed all the empty things. GTG!

jonathanprozzi (approved):
This looks good! Once we have some usage instructions we can update the Readme (same with all of our other things in Sage)

skuhlmann:
> This looks good! Once we have some usage instructions we can update the Readme (same with all of our other things in Sage)

Yeah I'll take a stab at that in the next PR that will have actual functionality.

Maybe we need a readme template?

Issue #68, opened 2022-04-18
Create Typography Components


jordanlesich:
# What 

- [x] Build all of the typography elements that are currently in the Figma doc.
- [x] Build as separate components
- [x] Build Mono Variant
- [x] Build Modifier spans, ex <Bold> <Black> <Italic> 
- Build stories that:
   - [x] display all the texts and which components to use
   - [x] give a rundown on how to use modifiers (with examples)

jonathanprozzi:
Closed by #74 🎉

PR #74, opened 2022-04-19
Feature/typography

jordanlesich:
## GitHub Issue
https://github.com/HausDAO/daohaus-monorepo/issues/68

## Changes

Creates typography components, modifier components, and stories for typography/modifiers. 

## Packages Added

NA

## Checks

- [x] App runs locally


libs/ui/src/components/atoms/typography/typography.tsx
```diff
@@ -0,0 +1,111 @@
...
+`;
+export const Italic = styled.span`
+  font-style: italic;
+`;
+export const Underline = styled.span`
+  text-decoration: underline;
+`;
+export const StrikeThrough = styled.span`
+  text-decoration: line-through;
+`;
```
alexkeating, line 111: I am wondering how things like Underline, Bold etc... are going to be used by consumers of the library. Is the expectation that they inherit those components in a new styled component and add other styles?

libs/ui/src/components/atoms/typography/typography.tsx
```diff
@@ -0,0 +1,111 @@
...
+  font-weight: ${font.weight.reg};
+  font-family: ${font.family.data};
+  letter-spacing: 1px;
+`;
+export const DataXl = styled.p`
+  font-size: ${font.size.xl};
+  font-weight: ${font.weight.reg};
+  font-family: ${font.family.data};
+  letter-spacing: 1px;
+`;
```
alexkeating, line 92: Should we merge these into a single component every other style is the same? Or possibly even inherit from a base Data component

libs/ui/src/components/atoms/typography/typography.tsx
jordanlesich: Not really, they're spans that you can use to nest inside of the text components, There's examples of how this could be used in the storybook. They can be used to modify individual words or characters instead of entire blocks of text. It's hard to target individual words with css and many linters don't like it when we nest <p>'s inside of <p>'s. 

In cases where all text within a given text element needs to change its font weight, we could: 

1) wrap the inner text with the modifier

or 

2) Ask why this text element isn't in the design system. Add it to the system, name and make the new element.

libs/ui/src/components/atoms/typography/typography.tsx
jordanlesich: I updated to so that they inherit. Definitely removed some code.

libs/ui/src/components/atoms/typography/typography.tsx
alexkeating: That makes sense, I was thinking more for like cases that are more complicated like bold underlined text. Would that look like:

```
<Underline>
  <Bold>
     Text
  </Bold>
</Underline>
```

libs/ui/src/components/atoms/typography/typography.tsx
jonathanprozzi: I think we should try to inherit shared base styles. This feels like it's pretty large. What is the reason for wanting to avoid using props? I'm open to it, but curious since I've always preferred things like size to be handled at the prop level. If we go with the components approach, I think we should try to inherit these. 

Edit: I saw that you updated to use inheritance.

libs/ui/src/theme/HausThemeContext.tsx
```diff
@@ -51,7 +51,7 @@ export const HausThemeProvider: FunctionComponent<{
   return (
     <HausThemeContext.Provider value={{ theme, setTheme, toggleLightDark }}>
       <ThemeProvider theme={theme}>
-        <GlobalStyles />
+        <GlobalStyles theme={theme} />
```
jonathanprozzi, line 54: Does this relate to what you mentioned about needing to pass in theme values into the GlobalStlyes?

jonathanprozzi (commented):
Looks good -- my questions are around mainly what @alexkeating already flagged about the choice of using components as opposed to props for sizes. I've always preferred props for this, so curious what you're thinking with it (I'm open either way as long as we're all good with it!)

libs/ui/src/components/atoms/typography/typography.tsx
jonathanprozzi: I'd imagine that would work given the structure, but I'm curious as well -- having these as `spans` makes a lot of sense so that we can have the flexibility to apply that to individual words.

libs/ui/src/components/atoms/typography/typography.stories.tsx
```diff
@@ -47,9 +47,7 @@ export const AllFonts = () => {
   return (
```
jonathanprozzi, line 47: One thing I'm wondering about (and this may be a question we have with Design as well) is how we'd want to use this in a case where we want to have styles separate from semantic meaning. For example, if we want to have something that is semantically an `<h3/>` given where it is in the HTML layout tree, but we want the style to be that of what we have set here as `<H1/>` -- one case may be where we have a heading on a Card. In the page layout, it may be an `<h3/>` but given the hierarchy on the Card we'd want it to be sized as what we have as an `<H1/>` 

I've handled this before by using the `as` prop that's been supported in certain libraries where you could do something such as `<Heading as='h3' size='xxl'/>`

This is something I'm wondering and thinking about just generally with how we're thinking about these Typographic elements @jordanlesich @alexkeating @earth2travis

libs/ui/src/theme/HausThemeContext.tsx
jordanlesich: Actually don't think this is needed.

libs/ui/src/components/atoms/typography/typography.tsx
jordanlesich: @alexkeating It works like that. If we are seeing a lot of bold, underlined words of (non-link) text, then we can make a new component.

libs/ui/src/components/atoms/typography/typography.stories.tsx
jordanlesich: Styled components supports the 'as' prop out of the box. So if we want a section text that is semantically different than the styling, we can edit it with that.

libs/ui/src/components/atoms/typography/typography.tsx
jordanlesich: Here's my reasoning for single component types. 

1) Simpler components. 
2) Easier to read
3) Faster to use, no typing in props. 
4) Limited functionality is a feature here. It enforces the styles out of the gate. 

All of these things fall in with line with the spec. I think we'd all had meetings about this too, right? The goal was to break down some of these primitives where we really don't need to be using props, but use single declarative props for things like button where props will always be a reality. 

Also worth noting that if we try this system and we really don't like it, we can compose all these components into larger Text components.

jordanlesich: Here's some additional reasoning that I think might help. 

```js 
const doAThing = (thing) => {
do(thing)
}

doAThing('thing1')
doAThing('thing2')
```

In the above example, we have a component that is maximally flexible and reusable. We could leave it as is. However we know that we're only going to do eight different things. The possibilities are finite,. 

In the spec, reliability, readability, and simplicity were key considerations for how this should be built. So does the function above best implement on our spec? 

```js
const thing1 = () => {
     doAThing('thing1')
}
const thing2 = () => {
     doAThing('thing2')
}
```
Instead of being maximally reusable or flexible, thing1() is better at being declarative, readable, and it's literally impossible to mess up. Also, the possibility narrowing has uses beyond fudging up the arguments , it constrains possibilities and helps ensure that we are in sync with the design system.  

Here's an additional consideration when it comes to DRY. Right now, on this PR, there's some extra code. However, we're not factoring all the needless props repeating that will happen in the future. I'd argue this is long-term more dry. 

A potential downside is that remembering to use <ParMd/> would be harder to remember to use <Text />. This is a true con. Also, it's a pattern that is less common. One component to handle them all would be easier to remember. But at the moment, I don't think the cons outweigh the pros.

alexkeating (approved):

jordanlesich:
Some conclusions that Keating and I reached on our call. We arrived on some general principles that I think are more important than what we do with our text elements. 

1) Keating mentioned that if we focus too deep on certain patterns in the beginning, we may end up in a situation where we 'code ourselves into a corner'. This is a pretty big problem given how we'd like to structure our new apps. Essentially, we implement a pattern everywhere, it's too much work to go update it, so we end up stuck. 

The conclusion that we should allow ourselves an 'escape hatch' so that we can always improve the code, without making breaking changes to applications. 

What is that escape hatch? 

2) Keating and I both noted that remembering that the system might be really handy for us, but could add unnecessary friction for community devs. While we do intend to serve our own interests first, we would also like to reduce friction for community devs in the long run. 

The escape hatch here is ensuring that we could always compose these elements into a more generic text element  in the future. If many of our community devs find it too hard to use typography as is, then we can build a generic text element that composes the atoms, while still supporting the primitives. By composing these elements we can offer the best of both worlds with minimal overhead. 

Being rigid with our styles is important for our apps. It might less important for the larger community. There's more than one way to skin a cat, our building blocks should be composable enough to accommodate that. In this case, I think they are. 

3) Readability is influenced by what people are used to. So in some cases, there is some merit to doing things the way other people do them. This should definitely be a consideration when designing code for others. In this case of this PR, I think the learning curve is pretty small and well worth it, but if it isn't we have options.

jonathanprozzi (approved):
Thanks for the added context and explanation.

Issue #23, opened 2022-04-01
Create Tooltip Component


skuhlmann:
# Overview

Create a reusable basic tool tip component.
- Single warning icon
- One color
- shows message onHover 

## Details

Name: `<ToolTip />`

### Spec questions

- Does Radix have an Icon component
    - this is a good chance to define how we do icons
-  Does Radix already have a tooltip?

#### Variants

- just looks like size.

#### Props
- toolTipText
- size

## Resources/Design

https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=87%3A2443

## Deliverable
- Code/PR
- Storybook deployment

jordanlesich:
Put some love into this component to hopefully prevent some of the challenges we had with tooltips before. 

Users can choose which side it renders on, whether they want to inject JSX or strings as content, and the trigger element. But the component is also customisation.  Devs can overwrite defaults with props. 
```js
const Tooltip = ({
  content = 'Your content here',
  side = 'right',
  triggerEl = <BiErrorCircle size="1.4rem" color={COLOR.WARNING} />,
}: {
  content: string | React.ReactNode;
  side?: 'top' | 'right' | 'bottom' | 'left';
  triggerEl: string | React.ReactNode;
}) => {
  return (
    <TooltipPrimitive.Provider delayDuration={400}>
      <TooltipRoot>
        <TooltipTrigger>{triggerEl}</TooltipTrigger>
        <StyledContent side={side} sideOffset={18}>
          {content}
          <StyledArrow />
        </StyledContent>
      </TooltipRoot>
    </TooltipPrimitive.Provider>
  );
};
```

jordanlesich:
For icons, I've been using them direct with React-Icons. So far, so good. 

As far as size goes, what part of size are we controlling? I just have it wrapping whatever gets input for ``content``, then it applies padding.

jonathanprozzi:
I think react-icons is a great call. One thing we can maybe do in our component library team meetings is to decide if there's a certain subset of react-icons we'd all want to use and align around that.

earth2travis:
Blocked by: https://github.com/HausDAO/daohaus-monorepo/issues/36

jordanlesich:
Figured it would be good to make a generic input that can be used anywhere instead. Then we can add it to the input as part of the InputLabel PR.

jonathanprozzi:
[...]

PR #77, opened 2022-04-21
Feature/tooltip

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/23

## Changes

- tooltip atom
- tooltip icon primitives
- tooltip variant stories (including custom icons and content) 

## Packages Added
"@radix-ui/react-tooltip": "^0.1.7"

## Checks

- [x] App runs locally


jordanlesich:
This merges the most recent changes on baseInput branch, which was mostly reviewed.

libs/ui/src/animations/general.ts
```diff
@@ -0,0 +1,19 @@
+import { keyframes } from 'styled-components';
```
jonathanprozzi, line 1: Love the idea of having an `animations` folder where we keep these.

libs/ui/src/components/atoms/input/input.stories.tsx
```diff
@@ -1,11 +1,57 @@
...
-export const InputAtom2 = () => <Input />;
-export const InputAtom3 = () => <Input />;
+const Template: ComponentStory<typeof Input> = (args) => <Input {...args} />;
+
+export const AtomInput = Template.bind({});
+
+AtomInput.args = {
+  placeholder: 'Placeholder',
+};
+export const InputWithIcon = Template.bind({});
```
jonathanprozzi, line 17: General question -- now that you've tried it, what do you think about using the `Template.bind` approach compared to the `export const` ?

libs/ui/src/components/atoms/input/style.ts
```diff
@@ -0,0 +1,73 @@
+import styled from 'styled-components';
+
+import { Theme } from '../../../types/theming';
```
jonathanprozzi, line 3: I ended up resolving this similarly locally -- we can sync about this later because we should be able to streamline the import a bit more

libs/ui/src/components/atoms/tooltip/types.ts
```diff
@@ -0,0 +1,7 @@
+export type TooltipProps = {
+  content?: string | React.ReactNode;
```
jonathanprozzi, line 2: Nice. I love this flexibility -- that's how I've structured types for `content` before

jonathanprozzi (approved):
This looks good to me! Runs and builds locally. Left a few comments throughout but this is good to go.

I noticed a lot of modifiers to the theme and types, I'm thinking those were modifications you made while building out the component? Makes sense, just curious!

libs/ui/src/components/atoms/input/input.stories.tsx
jordanlesich: Great. Storybook works much better with it.

libs/ui/src/components/atoms/input/input.tsx
```diff
@@ -1,3 +1,35 @@
...
+    long,
+    full,
+    icon,
+    warning,
+    error,
+    placeholder = 'Placeholder',
+    number,
+  } = props;
+
+  const inputClasses = classNames({ long, full, warning, error, number });
```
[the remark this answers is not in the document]
jordanlesich: I've been loving it so far.

alexkeating (approved):
Looks, good! Also, I think when we have these cascading prs it will be good to set the base branch to the branch that we branched off from to shrink the diff and also indicate the correct merge order. Because I merged in the input pr and it looks like it kicked up some conflicts with this pr

jordanlesich:
@alexkeating 

This is tricky. in both cases I started with develop as the base. Then, because of something I needed in some global type of style, I need to pull down the pending repo. Otherwise I'll get conflicts for those as well.

PR #79, opened 2022-04-21
Hotfix/typescript error

jordanlesich:
## GitHub Issue

No issue. Small type error. 

ThemeProvider inside of HausThemeContext was only expecting a single React Node as a return type, 

For some reason, I'd never seen the error until we installed the build locally on Adrienne and Avi's computers. 

I was able to reproduce the error by running storybook on my terminal instead of the VScode terminal. 

**Worth Noting**: This error was able to get past me running ``nx build ui`` and ``nx lint ui`` locally. It also made it past CI for a few PRs.  

## Changes

Wrapping the children inside of ThemeProvider in a ReactFragment fixes the error. 

## Packages Added
NA

## Checks

- [x] App runs locally


libs/ui/src/theme/HausThemeContext.tsx
```diff
@@ -20,22 +20,25 @@ type HausUI = {
   toggleLightDark: () => void;
 };
 
+type ProviderProps = {
+  children: ReactNode;
```
jonathanprozzi, line 24: I'm wondering if we want to outright allow for passing in multiple children in this since we're now adding the fragment. Something like:
`children: React.ReactNode|React.ReactNode[];`

@jordanlesich @alexkeating what ya'll think?

jonathanprozzi: I'm unsure how React handles this -- I'm wondering if the Fragment is treated as a single ReactNode child (and if so, then the Array is unnecessary in the type)

jonathanprozzi (approved):
This looks good to me, still curious why this happened though. I'm wondering if we want to add the enum to allow for an array of children (see comment) or if this isn't necessary if React treats everything in a Fragment as a single ReactNode. Either way this looks good to me. I still want to dig in a bit more about the error itself though.

Issue #26, opened 2022-04-01
Create Input Message Component


skuhlmann:
# Overview

Create a reusable basic input message component.

## Details

Name: `<InputMessage />``

### Spec questions

- Dependent on some Typography/icons

#### Variants

https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=92%3A8758

- warning w/ icon
- valid value w/ icon
- error w/ icon

#### Props
- messageText
- messageType (warn, error, valid)

## Resources/Design

PL Page: https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=92%3A8758

## Deliverable
- Code/PR
- Storybook deployment

earth2travis:
Blocked by: https://github.com/HausDAO/daohaus-monorepo/issues/36

jordanlesich:
Changes to the spec. 

This was written a while back and we've made some changes since then. Most of the logic now happens in the FieldWrapper component, also we have new considerations like theming and dark/light mode. 

this spec will be more about building a bunch of the input primitives. The field wrapper spec will handle their logic and states. 

Here's what this spec should include
- [ ] input atom (with standard state changes)
- [ ] input atom border state change (error, warning, etc)
- [ ] add on variants (excluding button)
- [ ] input consumes color values from theme + handles light/dark shifts

The FieldWrapper is what composes the helperText.

 I'll build message text as a separate story and include it on the PR for this issue, but won't integrate it as a macro until

PR #82, opened 2022-04-22
Feature/helper texts

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/26

## Changes

- Generic helper text
- Specialised helper texts w/icons
- Stories for helper texts
- All texts now accept a color prop
- Reorg component folder structure as per earlier discussion 

## Packages Added
na

## Checks

- [x] App runs locally


libs/ui/src/components/atoms/helperTexts/helperTexts.tsx
```diff
@@ -0,0 +1,59 @@
...
+export type HelperTextType = {
+  color: string;
+  icon?: IconType;
+  children: React.ReactNode;
+};
+export type SpecialHelperText = {
+  children: React.ReactNode;
+};
+
+export const HelperText = ({ color, icon, children }: HelperTextType) => {
```
alexkeating, line 19: I guess ultimately size will be a prop for these

alexkeating (approved):

libs/ui/src/components/atoms/helperTexts/helperTexts.tsx
jordanlesich: I don't think so...I'm assuming all input helper texts will be the same. If not we can pass in a text that we'd like to use. Similar to what we do with the icon.

alexkeating: Okay, good to know thanks

Issue #25, opened 2022-04-01
Create Input Wrapper Component


skuhlmann:
# Overview

Molecule component wrapping all the atoms needed for an input.

## Details

Name: `<InputWrapper />`

### Spec questions

-  How does this play with form builder/react hook form and what do we need to consider around that?

#### Atoms included
- `<Input />`
- `<Label />`
- `<ToolTip />`
- `<InputMessageText />`

#### Variants

https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=92%3A8758

- Basic input:
    - Filled
    - With placeholder
    - Empty
    - w/ focus (outline)
- w/ Button
- w/ icon
- w/ Label
- w/ Tooltip
- Full length address display
- w/ valid response message
- w/ warning
- w/ error
- w/ required indicator

#### Props
- placeholder
- key
- id
- name
- onChange
- placeholder
- ref
- disabled
- defaultValue
- appendComponent
    - Takes a component (like a button or a select) to append after the input 
    - https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=92%3A8758
- required
- warning message
- error message
- valid repsonse message
- isAddress

## Resources/Design

PL Page: https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=92%3A8758

## Deliverable
- Code/PR
- Storybook deployment

PR #86, opened 2022-04-23
Feature/field wrapper

jordanlesich:
## GitHub Issue
https://github.com/HausDAO/daohaus-monorepo/issues/25

## Changes

- Helper text factory
- Input label component
- address variant 
- slight story file restructure 
- clean up

## Packages Added
na

## Checks

- [x] App runs locally


libs/ui/src/components/atoms/helperTexts/helperTexts.tsx
```diff
@@ -57,3 +67,17 @@ export const ErrorText = ({ children }: SpecialHelperText) => {
     </HelperText>
   );
 };
+
+export const HelperTextFactory = ({
+  error,
+  success,
+  warning,
+  helperText,
```
alexkeating, line 75: Since these arguments are mutually exclusive it will be cleaner to have two arguments. One for text and then another for type of text.

libs/ui/src/components/atoms/helperTexts/helperTexts.tsx
```diff
@@ -57,3 +67,17 @@ export const ErrorText = ({ children }: SpecialHelperText) => {
...
+}: HelperTextFactoryProps) => {
+  if (!error && !success && !warning && !helperText) return null;
```
alexkeating, line 77: Once the above params are change we can do this check on an object of values rather than having a long boolean statement

libs/ui/src/components/atoms/helperTexts/helperTexts.tsx
alexkeating: Text could even be changed to children potentially to make this a single argument function

jordanlesich: Generally agree with this, but this component is trying to solve a different problem. Design had mentioned that if we have an input that has more than one helpertext on it, instead of displaying both, we render only one based on a certain priority. 

Inputting one element would require that we do similar logic (taking possibly more than one type of helper text, but displaying only one) in the component above.

jordanlesich: I guess it's a sort of reducer at this point. Could change the name from factory to reducer.

alexkeating (approved):

Issue #66, opened 2022-04-18
Scaffold the Sage Hub App


jonathanprozzi:
## What

Scaffold the Hub App Sage package in the monorepo. We can add as much as we're able to to this package while Design is finishing up the v1 for the UI.

## TODOs

- [ ] Create the hub-app package 
- [ ] Begin adding any scaffolding that isn't reliant on Design

## Deliverables

- PR with the `hub-app` added to Sage.

PR #87, opened 2022-04-25
initial hub

alexkeating:
## GitHub Issue

#66 

## Changes

Generate initial hub

## Checks

- [ ] App runs locally


jordanlesich (commented):
Seems pretty straight forward to me.

apps/hub-app/src/main.tsx
```diff
@@ -0,0 +1,14 @@
+import { StrictMode } from 'react';
+import * as ReactDOM from 'react-dom';
+import { BrowserRouter } from 'react-router-dom';
+
+import App from './app/App';
+
+ReactDOM.render(
+  <StrictMode>
+    <BrowserRouter>
```
skuhlmann, line 9: We should probably do some tests this week with skynet deployments and routing. We might need to switch to HashRouter to enable deep linking into routes. This was the case with other ipfs deployments in the past.

skuhlmann (approved):

apps/hub-app/src/main.tsx
alexkeating: Yes, we will need to use hash router. Skynet doesn't support browser router

PR #102, opened 2022-04-27
scaffold button, class generation

alexkeating:
primary, secondary, outline styles

large and small styles

add base button story for audit

change name 'outline' to tertiary

overhaul button theming

add disabled stateto stories

click handler

## GitHub Issue
https://github.com/HausDAO/daohaus-monorepo/issues/96

## Changes

- This PR creates buttons with all CSS for each variant. 

## Packages Added

NA

## Checks

- [x] App runs locally


libs/ui/src/components/atoms/button/button.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+type ButtonProps = {
+  children: React.ReactNode;
+  secondary: boolean;
+  sm: boolean;
+  lg: boolean;
+  tertiary: boolean;
+  disabled: boolean;
+  onClick?: React.MouseEventHandler<HTMLButtonElement>;
+};
```
alexkeating, line 13: Should some of these be combined? We can't have a lg and small button or a secondary and tertiary button correct?

jordanlesich: Right. This follows the spec though in keeping the props minimalistic booleans. 

I think we could trust devs to not set two styles or colors at once. Also worth mentioning that we could possibly double set without this pattern as well.

jordanlesich (approved):

Issue #113, opened 2022-04-30
Create `daohaus-connect` UI States in Storybook


jordanlesich:
# What: 

Build out all the UI combinations for each state of the dao-haus connect component. 

https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=4005%3A28979

To build
- [x] Warning button (ex. Network Unavailable)
- [x] Avatar button 
- [x] connect button 
- [x] wrong network dropdown 
- [x] network unavailable dropdown
- [x] Connected dropdown

Find out: 
- [x] What happens when we're on the right network. 

Make New Issues: 

- [x] Etherscan link molecule (works in every app. Takes in current network, address, and blob of scan links.

jordanlesich:
Removed iconbutton component as I'm not sure if we're actually doing that yet (in terms of the design system) and the explorer icon task is a link, not a button.  

However, we will need a generic Avatar component.

jonathanprozzi:
This is closed by #134

PR #134, opened 2022-05-05
Feature/daohaus connect base UI

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/113
## Changes

Listed in issue. 

## Packages Added

 "@radix-ui/react-avatar": "^0.1.4",

## Checks

- [x] App runs locally


jordanlesich:
Hey, I just want to mention that I tried a new pattern/workflow for this that didn't really work. 

My goal was to create all the compound elements in a story and compose them together into all the UI state needed for daohaus-connect and then temporarily add them to Storybook as a set of recipes. 

Doing so would allow me to focus on the UI and refining the atoms before scaffolding out a new package and building there. 

Here's what went wrong:  

- I want to compose base molecules inside of a story. This doesn't jive well with the .bind({}) pattern and in more complex cases  and will cause stack overflows. 
- So I tried the classic way, by simply exporting custom molecules. This also caused stack overflows. 

The result is that I had to use a bit of both patterns. It's a messy story, but I have absolutely no desire to fix this up as all of this is just temporary anyways. 

So let's review with the understanding that once I build out daohaus-connect, DaoHausConnect.stories.tsx will no longer exist.

jonathanprozzi:
Thanks for the notes Jord -- I'll take a look at this a bit later tonight.

libs/ui/src/theme/global/mediaQueries.ts
```diff
@@ -0,0 +1,19 @@
+export const widthBreakpoint = {
+  mobileSm: 350,
+  mobileMd: 378,
```
alexkeating, line 3: Will we really have css changes between small and medium mobile? Seems like one could be eliminated but I am fine either way

jonathanprozzi: I think that's a good point and in actuality it's probably unlikely, but may not hurt to have the extra breakpoints in there incase it does become a distinction for a feature later on

jonathanprozzi (approved):
Lookin good!

I agree about the breakpoints, but I think it's okay to have the granularity there for future use. If we truly don't use the `md` we can always remove it

jordanlesich:
I also agree about the breakpoints. I just lifted it from one of my other apps. I'll meet with design and see where they are with breakpoints.

Issue #141, opened 2022-05-09
Resolve Jest issue with Ceramic


jonathanprozzi:
## Overview

This is to continue to explore the Jest and Ceramic issue to unblock #92.

## Details

There is an issue between Jest and Ceramic that is preventing the build and is blocking adding Ceramic to `haus-data`

- Converstation with the Ceramic team: https://discord.com/channels/682786569857662976/875762106245464125/971519976047513630

- Created this card so that the work being done to resolve this is represented

jonathanprozzi:
Closed by #146

PR #146, opened 2022-05-11
Fix jest issue

alexkeating:
## GitHub Issue

#141 

## Changes

- Change jest configuration to reflect ceramic core
- Add ceramic data to profile

## Packages Added

- @swc/jest

## Checks

- [x] App runs locally


.github/workflows/ci_master.yaml
```diff
@@ -24,12 +24,13 @@ jobs:
         run: |
           touch .env
           echo NX_RIVET_KEY=${{ secrets.RIVET_KEY }}  >> .env
+          echo CERAMIC_NETWORK=${{ secrets.CERAMIC_NETWORK }}  >> .env
```
jonathanprozzi, line 27: I'd love a brief TLDR at some point when you have bandwidth of what you had to do to fix this -- it looks like you had to add a specific CERAMIC_NETWORK key for this to work?

Mainly just curious!

libs/dao-data/test/profile.spec.ts
```diff
@@ -22,4 +22,12 @@ describe('haus', () => {
     );
     expect(profile.ens).toBe('samkuhlmann.eth');
```
jonathanprozzi, line 23: I wonder if we should try to set up a profile for a DAOhaus ENS or something so we don't have to use Sam's in the future?

jonathanprozzi (approved):
This is great -- awesome work @alexkeating 
🔥🔥🔥

.github/workflows/ci_master.yaml
alexkeating: Yes, definitely

libs/dao-data/test/profile.spec.ts
alexkeating: Yes, agreed

PR #150, opened 2022-05-13
Fix issues

alexkeating:
## Changes

Fix caip issue

## Packages Added
"@ceramicnetwork/3id-did-resolver": "^2.0.1-rc.4",
"@ceramicnetwork/http-client": "2.0.1-rc.4",
"@ceramicnetwork/stream-caip10-link": "2.0.1-rc.2",
"@ceramicnetwork/stream-tile": "2.1.0-rc.3",
"@gnosis.pm/safe-apps-web3modal": "^12.0.0",

## Checks

- [ ] App runs locally


libs/dao-data/src/Profile.ts
```diff
@@ -66,3 +77,59 @@ export default class Profile {
     return await provider.lookupAddress(address);
   }
 }
+
+const profilModel = {
```
skuhlmann, line 81: Do we need this here?

alexkeating: We can remove it. It has the metadata of for the hash we use above. I don't mind either way

libs/dao-data/src/Profile.ts
```diff
@@ -1,5 +1,7 @@
-import { Core } from '@self.id/core';
+// import { Core } from '@self.id/core';
```
jordanlesich, line 1: comment

libs/dao-data/src/Profile.ts
```diff
@@ -25,12 +27,11 @@ type AccountProfile = {
 const CERAMIC_NETWORK = process.env['CERAMIC_NETWORK'] || '';
```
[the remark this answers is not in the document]
alexkeating: Yes, good catch

libs/dao-data/src/Profile.ts
```diff
@@ -25,12 +27,11 @@ type AccountProfile = {
...
 
 export default class Profile {
-  client: Core;
+  // client: Core;
```
jordanlesich, line 30: comment

libs/dao-data/src/Profile.ts
```diff
@@ -66,3 +77,59 @@ export default class Profile {
     return await provider.lookupAddress(address);
   }
 }
+
+const profilModel = {
+  definitions: {
```
jordanlesich, line 82: This is a pretty massive blob of data. Would it make sense to refactor to increase readability?

skuhlmann (approved):

PR #152, opened 2022-05-13
Remove process from deps

alexkeating:
## GitHub Issue

- None 

## Changes

Fix process missing issues when importing haus

## Packages Added

## Checks

- [ ] App runs locally

### DIFF
--- a/libs/dao-data/src/Profile.ts
+++ b/libs/dao-data/src/Profile.ts
@@ -23,13 +23,12 @@ type AccountProfile = {
   background?: string;
 };
 
-const CERAMIC_NODE = process.env['CERAMIC_NODE'] || '';
-
 export default class Profile {
   providers: Keychain;
 
-  constructor(providers: Keychain) {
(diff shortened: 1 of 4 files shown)


libs/dao-data/src/Profile.ts
```diff
@@ -23,13 +23,12 @@ type AccountProfile = {
   background?: string;
 };
 
-const CERAMIC_NODE = process.env['CERAMIC_NODE'] || '';
-
 export default class Profile {
   providers: Keychain;
 
-  constructor(providers: Keychain) {
+  constructor(providers: Keychain, node?: string) {
```
skuhlmann, line 29: Do we also need to update the Haus.ts create function to pass this in?

skuhlmann: I can handle this next week actually. I want to play with it.

alexkeating: Yea, good point

skuhlmann (approved):

Issue #165, opened 2022-05-17
[DATA] Profile Fetch Bug


skuhlmann:
# Description

- Haus.profile.get() crashes if we pass in an address w/out a ceramic
- Catch error and provide fallbacks

alexkeating:
What is the error?

skuhlmann:
> What is the error?

`TypeError: Cannot use 'in' operator to search for '_tag' in null`

something inside that call to TileDocument.load

PR #167, opened 2022-05-17
Feature/refactor error handling

skuhlmann:
## GitHub Issue

Add a link to the GitHub issue.
https://github.com/HausDAO/daohaus-monorepo/issues/165

## Changes

- Fixes the profile.get failure on an address without a profile
- Refactors error class and applies in the queries class functions
- Clean sup old unused stuff

## Packages Added

n/a

## Checks

- [x] App runs locally


libs/dao-data/src/Profile.ts
```diff
@@ -66,14 +46,23 @@ export default class Profile {
...
   private async getEns(address: string): Promise<string | null> {
     const provider = new ethers.providers.JsonRpcProvider(
       this.providers['0x1']
     );
-    return await provider.lookupAddress(address);
+    try {
+      return await provider.lookupAddress(address);
+    } catch {
+      return null;
+    }
```
alexkeating, line 66: So, this is another sort of philosophical thing. Do we hide errors in the library or do we allow clients to handle errors. Personally, I am a fan of the later

alexkeating (approved):
- Left one comment that may be worth a discussion

libs/dao-data/src/Profile.ts
skuhlmann: Yeah worth a discussion.

The spot above was spitting out some bad errors on the Ceramic fetch if `defId` was null in the case the address didn't have a profile. So that seemed like an error in our library not due to what the client is doing.

In the case of profile I kind feel we ought to just return an empty profile + ens if that exists... but I see what you mean.

alexkeating: Yea true, in general this can cause big headaches in debugging issues, because we are essentially swallowing the error here.

skuhlmann: Good point. Adding a ticket to revisit error handling in general for the SDK. 

I've taken a few stabs at it in the Query class and add an error message in the response. But could look more into Throwing as a pattern.

In this case It seems ceramic or our implementation of it was more of a problem. If they throw an error on the missing profile then we kinda have to catch that for our use case. For us a missing profile isn't an error.

PR #177, opened 2022-05-19
Feature/avatar

jordanlesich:
## GitHub Issue

https://github.com/orgs/HausDAO/projects/3/views/3

## Changes

- Creates Avatar atom 
- Atom stories
- Creates ProfileAvatar molecule
- Molecule stories
- Hook up avatar to daohaus-connect

## Packages Added

    "ethereum-blockies-base64": "^1.0.2",

## Checks

- [x] App runs locally


jordanlesich:
Note: due to the format we receive from Haus.profile ProfileAvatar component will not be able render ceramic images. 

More discussion here: 

https://discord.com/channels/709210493549674598/951186621397618708/976593804448632873

libs/ui/src/components/molecules/ProfileAvatar/ProfileAvatar.tsx
```diff
@@ -0,0 +1,30 @@
...
+  address?: string;
+  image?: string;
+};
+
+export const ProfileAvatar = ({
+  address,
+  image,
+  ...props
+}: ProfileAvatarProps) => {
+  const blockie = useMemo(() => {
```
jonathanprozzi, line 16: Ah good idea to memoize this!

jonathanprozzi (approved):
Looks good to me -- did you have a chance to sync up with @alexkeating about the use within Hub? If not we can all jam during Show & Tell later maybe

jonathanprozzi:
I think we may be having conflicts from the InputWrapper related changes -- wondering why those were included in this? Did something need to be changed while making the Avatar? Just curious -- do you think we should merge this PRs in a specific order because of this (I believe #176 also has some changes to this?)

0xRowdy:
> I think we may be having conflicts from the InputWrapper related changes -- wondering why those changes were included in this? Did something need to be changed while making the Avatar? Just curious -- do you think we should merge this PRs in a specific order because of this (I believe #176 also has some changes to this?)

Could have been related to re-naming the folder from InputWrapper to FieldWrapper. That was only done to keep the consistancy with the actual component name.

alexkeating (approved):

PR #182, opened 2022-05-20
Refactor/network and rpc as props

jordanlesich:
## GitHub Issue

https://github.com/orgs/HausDAO/projects/3/views/3?filterQuery=label%3A%22daohaus-connect%22

## Changes

- Enables community developers to add their own rpc vars
- Uses ours by default
- Since RPC key are stored in .env community devs will not have access to our keys. 
- Tested narrowing networks through props.  

## Packages Added

NA

## Checks

- [x] App runs locally

### DIFF
--- a/apps/summon-app/src/app/constants.ts
+++ b/apps/summon-app/src/app/constants.ts
@@ -1,0 +1,27 @@
+// FOR DEMONSTRATIOn
+export const limitedNetworkTest = {
+  '0x1': {
+    chainId: '0x1',
+    networkId: 1,
+    name: 'Mainnet',
+    symbol: 'ETH',
+    rpc: `https://${import.meta.env.VITE_RIVET_KEY}.eth.rpc.rivet.cloud/`,
+    explorer: 'https://etherscan.io',
+  },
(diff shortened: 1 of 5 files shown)


apps/summon-app/src/app/constants.ts
```diff
@@ -0,0 +1,27 @@
+// FOR DEMONSTRATIOn
```
jonathanprozzi, line 1: Makes sense to include this! We can add to some docs (I'll add to the list) explaining this and the env vars that are needed (and how to set them up)

apps/summon-app/src/main.tsx
```diff
@@ -5,6 +5,8 @@ import * as ReactDOM from 'react-dom';
 import { BrowserRouter } from 'react-router-dom';
 
 import App from './app/App';
+// FOR DEMONSTRATIOn
+// import { limitedNetworkTest } from './app/constants';
```
jonathanprozzi, line 9: Is this showing how this would be imported? Sorry if that's a really straight forward question ha

Issue #47, opened 2022-04-14
@daohaus/data: Add functions/pattern to hydrate response data on scope queries


skuhlmann:
## Overview

Add a system for hydration data to add/transform before returning the response. 

### Details

1st thing to implement here is tto add proposalStatus to the Proposal entity. Early example:
https://github.com/HausDAO/daohaus-rage/blob/master/libs/data/src/utils/hydrators.ts

The system should be able to support:
- Any function that transforms or adds fields
- Async fetch data from other sources to add into the main entity
- Consider efficiency when hydrating a large list of entites

skuhlmann:
https://github.com/HausDAO/daohaus-monorepo/pull/190

PR #190, opened 2022-05-23
Feature/data transformers

skuhlmann:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/47

## Changes

- [x] transformer functions added for proposals, tokenbalances and profile
- [x] use entity types generated off graphql schema

## Packages Added

n/a

## Checks

- [x] App runs locally


skuhlmann:
@alexkeating The issue with circular dependencies can be seen in the last commit.

Basically I have a help function i was going to move to the commons package, but it relied on a type that lives in dao-data because it's part of the gql generated types. Not a big deal to have that helper function in the dao-data package (that's how i fixed it), but could be worth moving Queries and types into the common package.

libs/dao-data/src/Query.ts
```diff
@@ -110,7 +118,8 @@ export default class Query {
     },
     filter,
   }: ListQueryArguments<Proposal_OrderBy, Proposal_Filter>): Promise<
-    QueryResult<ListProposalsQuery>
+    QueryResult<TransformedProposalListQuery>
+    // QueryResult<ListProposalsQuery>
```
jordanlesich, line 122: Do we need the comment?

libs/dao-data/src/subgraph/queries/daos.generated.ts
```diff
@@ -10,14 +10,14 @@ export type ListDaosQueryVariables = Types.Exact<{
 }>;
 
 
-export type ListDaosQuery = { daos: Array<{ id: string, createdAt: string, transactionHashSummon: string, lootAddress: string, safeAddress: string, lootPaused: boolean, sharesPaused: boolean, gracePeriod: string, votingPeriod: string, proposalOffering: string, quorumPercent: string, sponsorThreshold: string, minRetentionPercent: string, shareTokenName?: string | undefined, shareTokenSymbol?: string | undefined, totalShares: string, totalLoot: string, latestSponsoredProposalId: string, metaData?: { name?: string | undefined } | undefined }> };
+export type ListDaosQuery = { daos: Array<{ id: string, createdAt: string, transactionHashSummon: string, lootAddress: string, safeAddress: string, lootPaused: boolean, sharesPaused: boolean, gracePeriod: string, votingPeriod: string, proposalOffering: string, quorumPercent: string, sponsorThreshold: string, minRetentionPercent: string, shareTokenName: string | undefined, shareTokenSymbol: string | undefined, totalShares: string, totalLoot: string, latestSponsoredProposalId: string, metaData: { name: string | undefined } | undefined }> };
```
jordanlesich, line 13: format

jordanlesich: For whole doc here

libs/dao-data/src/subgraph/queries/members.generated.ts
```diff
@@ -10,14 +10,14 @@ export type ListMembersQueryVariables = Types.Exact<{
 }>;
 
 
-export type ListMembersQuery = { members: Array<{ id: string, createdAt: string, memberAddress: string, shares: string, loot: string, delegatingTo: string, delegateShares: string, votes?: Array<{ createdAt: string, approved: boolean, balance: string }> | undefined }> };
+export type ListMembersQuery = { members: Array<{ id: string, createdAt: string, memberAddress: string, shares: string, loot: string, delegatingTo: string, delegateShares: string, votes: Array<{ createdAt: string, approved: boolean, balance: string }> | undefined }> };
```
jordanlesich, line 13: Format

libs/dao-data/src/utils/proposals.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (proposal.passed) {
+    return PROPOSAL_STATUSES.passed;
+  }
+  if (proposal.actionFailed) {
+    return PROPOSAL_STATUSES.actionFailed;
+  }
+
+  const now = new Date().getTime() / 1000;
+  if (
+    Number(proposal.votingStarts) < now &&
```
jordanlesich, line 20: Refactoring all these 'if' statements to functions (ex. if(isVotingPeriod(proposal)) ) will: 

- Make this more readable
-  Allow us to reuse this functionality later on in the UI
- Ensure that the methods we use to get proposal status are consistent across all of our apps.

libs/dao-data/src/utils/transformers.ts
```diff
@@ -1,6 +1,57 @@
...
+  ens: string | null,
+  profile: BasicProfile
+): AccountProfile => {
+  return {
+    address,
+    ens,
+    ...profile,
+    image:
+      profile.image?.original?.src &&
+      `https://daohaus.mypinata.cloud/ipfs/${profile.image.original.src.match(
```
jordanlesich, line 35: I'm gathering that this match() statement is a way to derive a Pinata enpoint from the original IPFS hash. 

Could we refactor this so it could be used again when we handle other non-profile images?

jordanlesich (commented):
Looking slick. 

Left a bunch of comments about refactoring some elements to utilities so they can be used later.

libs/dao-data/src/Query.ts
```diff
@@ -121,16 +130,27 @@ export default class Query {
...
+        ListProposalsQueryVariables
+      >(ListProposalsDocument, url, networkId, {
+        where: filter,
+        orderBy: ordering.orderBy,
+        orderDirection: ordering.orderDirection,
+      });
+
+      const proposals = queryResult.data?.proposals
+        ? queryResult.data?.proposals.map((prop) =>
+            transformProposal(prop as Partial<Proposal>)
```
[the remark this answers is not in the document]
skuhlmann: > Curious why Partial. Shouldn't the shape of proposals data all generally be the same?

The Type is generated from the subgraph schema, but the query doesn't pull all potential fields and subfields. All I've seen so far recommends using Partial.

libs/dao-data/src/utils/proposals.ts
skuhlmann: I like that!

libs/dao-data/src/utils/transformers.ts
skuhlmann: > I'm gathering that this match() statement is a way to derive a Pinata enpoint from the original IPFS hash.
> 
> Could we refactor this so it could be used again when we handle other non-profile images?

Yeah I like that. Going to add a future refactor ticket here to address once we get another use case/image source.

skuhlmann:
Refactor ticket coming out of this:
https://github.com/HausDAO/daohaus-monorepo/issues/200

jordanlesich (approved):

Issue #189, opened 2022-05-23
Set up Hub and Summoner build and Skynet deployment


skuhlmann:
## Overview

This issue is for setting up the Hub and Summoner build processes and CI/CD to deploy both frontends to Skynet.

## Details

- We can split this into 2 separate tickets if this is too big, but these are likely similar solutions for each app since they're both set up in the same way
- We want to ensure that our `nx build` is properly compiling TypeScript each time we build, and we also need to set up a GitHub Action so that we have CI/CD for building and deploying each app frontend to Skynet
  - Deploy to Skynet on each merge to `master` that passes tests and properly builds 

## Deliverables

- [ ] Get Hub and Summoner properly compiling TypeScript on build command
- [ ] Set up GitHub Actions deployment scripts to Skynet

PR #214, opened 2022-05-30
Bug/typecheck apps

alexkeating:
## GitHub Issue

#189 

## Changes

Add typecheck step for existing apps and deployment for hub and summon

## Packages Added

- nx-skynet

## Checks

- [ ] App runs locally

### DIFF
--- a/apps/hub-app/src/components/Profile.tsx
+++ b/apps/hub-app/src/components/Profile.tsx
@@ -54,7 +54,7 @@ const Profile = () => {
           image: profile.image || '',
           description: profile.description || '',
           emoji: profile.emoji || '',
-          background: profile.background || '',
+          background: profile.image || '',
         });
       }
     } catch (err) {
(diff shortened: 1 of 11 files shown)


apps/hub-app/src/components/Profile.tsx
```diff
@@ -54,7 +54,7 @@ const Profile = () => {
           image: profile.image || '',
           description: profile.description || '',
           emoji: profile.emoji || '',
-          background: profile.background || '',
+          background: profile.image || '',
```
skuhlmann, line 57: FYI - I've removed this field from the SDK. Looks like it was removed from the designs, right?

skuhlmann (approved):

apps/hub-app/src/components/Profile.tsx
alexkeating: Ah no worries, going forward this pr should raise those issues as the changes are made.

Issue #169, opened 2022-05-18
Get Wallet Connect working with daohaus-connect


skuhlmann:

jonathanprozzi:
This is a vite specific issue I believe. @alexkeating do you think that the approach for defining global in the index.html may not work since this is within a package?

alexkeating:
https://github.com/raid-guild/headline/blob/master/index.html#L19

PR #215, opened 2022-05-30
Support wallet connect

alexkeating:
## GitHub Issue

#169 

## Changes

Add global to app

## Checks

- [ ] App runs locally

### DIFF
--- a/apps/summon-app/index.html
+++ b/apps/summon-app/index.html
@@ -8,6 +8,11 @@
   </head>
   <body>
     <div id="root"></div>
-    <script type="module" src="/src/main.tsx"></script>
+    <script>
+      window.global = window;
+    </script>
+    <script type="module" src="/src/main.tsx">
+      window.global = window;
+    </script>
   </body>
 </html>
--- a/libs/daohaus-connect-feature/src/utils/defaults.ts
+++ b/libs/daohaus-connect-feature/src/utils/defaults.ts
@@ -10,7 +10,6 @@ export const supportedNetworks = addKeychain(
 );
 
 const providerOptions: IProviderOptions = {
-  // TODO
   walletconnect: {
     package: WalletConnectProvider,
     options: {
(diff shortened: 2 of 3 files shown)


apps/summon-app/index.html
```diff
@@ -8,6 +8,11 @@
   </head>
   <body>
     <div id="root"></div>
-    <script type="module" src="/src/main.tsx"></script>
+    <script>
```
skuhlmann, line 11: The only thing that worries me about this is that we need all clients using this package to add this to their app UI.

If this is the only way to solve - then maybe we need to get some 'Getting Started' docs for this package and ensure this is communicated. Thoughts @jonathanprozzi ?

libs/daohaus-connect-feature/src/utils/defaults.ts
```diff
@@ -11,16 +11,15 @@ export const supportedNetworks = addKeychain(
 
 const providerOptions: IProviderOptions = {
   // TODO
```
skuhlmann, line 13: TO-DONE

alexkeating: Updated

apps/summon-app/index.html
alexkeating: Yes, it will be necessary for all vite apps. We should document it. Let me know where and I can add something

skuhlmann: oh ok. just for vite. cool that alleviate my concern.

skuhlmann (approved):

apps/summon-app/index.html
jonathanprozzi: I was wondering that too since this is within our package. We can be sure to add this into the docs

jonathanprozzi (approved):
Nice. Wonder if we want to add the docs to our Wiki or to the package README? I can add to the README first and the Wiki but I'm wondering at what point we may also want a dedicated dev docs site. Something for us to keep on our radar @skuhlmann @alexkeating @earth2travis

Issue #143, opened 2022-05-10
Update Color System + Global Styles (summon)


adriennetilley:
Inputs:
https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=92%3A8758 

Buttons:
https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=67%3A1122

Checkboxes, radios, switches:
https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=44%3A2049

For examples of how these work on light mode:
https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=4320%3A63083

Spec includes styles for:
- Input color and states
- Button color and states 
- Standard link style and hover 

Notes:
- The app background for all apps will be Slate/Dark/1 or Slate/Light/1 for light mode
- Each app will have a custom primary (main button actions) and secondary (input backgrounds, default buttons) colors, these current choices are specific to the summon app

earth2travis:
@0xRowdy

PR #241, opened 2022-06-06
Feature/color and global styles

0xRowdy:
## GitHub Issue

[GitHub issue.](https://github.com/HausDAO/daohaus-monorepo/issues/143)

## Changes

Re-organize the styles and theme. Create a colors file, and create aliases for colors that map to utility to hold colors based on Radix Colors.

## Packages Added

NA

## Checks

- [x] App runs locally


libs/ui/src/components/atoms/Switch/Switch.tsx
```diff
@@ -73,7 +73,7 @@ export const Switch = React.forwardRef(
           onCheckedChange={onCheckedChange}
           disabled={disabled}
         >
-          <SwitchSlider className={classes} ref={ref} />
+          <SwitchSlider ref={ref} />
         </SwitchBase>
         <LabelContainer>
           <Label id={id} className={classes}>
```
0xRowdy, line 79: We should make a ticket to refactor Switch to use Label Component. 
I don't want to keep bloating this ticket.

libs/ui/src/theme/theme.ts
```diff
@@ -1,167 +1,226 @@
...
   PrimaryLightButton,
   SecondaryDarkButton,
   SecondaryLightButton,
   TertiaryDarkButton,
   TertiaryLightButton,
-} from './component/button';
+} from './atoms/button';
+import { FieldTheme, FieldThemeDark } from './atoms/field';
 
 export const defaultDarkTheme: Theme = {
```
0xRowdy, line 24: I would be interested in the idea of creating an atomic design folder structure for the theme?
I'll put togeather a hack md file to cover the pros & cons.

libs/ui/src/components/atoms/Switch/Switch.tsx
jordanlesich: Agreed.

libs/ui/src/theme/theme.ts
jordanlesich: I like the rigour, but we're in ultra-budget mode now. Probably best to leave it for now.

libs/ui/src/components/atoms/Button/Button.styles.ts
```diff
@@ -6,142 +6,118 @@ export const ButtonBase = styled.button`
   background-color: ${({ theme }) => theme.button.primary.bg};
   border: 0.1rem solid
     ${({ theme }: { theme: Theme }) => theme.button.primary.border};
+  border-radius: 0.4rem;
```
jordanlesich, line 9: Don't we have a const for this?

libs/ui/src/components/atoms/Button/Button.styles.ts
```diff
@@ -152,27 +128,33 @@ export const ButtonBase = styled.button`
...
       height: 1.5rem;
       margin-right: 0.8rem;
+      width: 1.5rem;
     }
   }
+
   svg.icon-left {
     margin-right: 1rem;
   }
+
```
jordanlesich, line 157: This component looks much cleaner

jordanlesich (approved):
This is looking awesome. Super thorough. There's lots of work here. Thanks man!

jordanlesich:
I changed the tooltip color as I noticed that it was secondary in the designs. I also added a generic secondary to the main theme. I went through everything in Storybook. Looking good.

Issue #60, opened 2022-04-18
Summoner Form Rules and Validation


jonathanprozzi:
## Overview

Create the rules and validation schema for the Summoner Form. This supports other user stories and scenarios so this ticket is lighter on those details.

## Deliverables

- Summoner form hooked up to React Hook Form
- Validation for each field including the following:
  - Required fields
  - Custom validation (such as addresses)
  - Leverage `yup` for other common validations such as strings
- Summoner validation created and integrated

## Detail

- Integrate with React Hook Form and look into leveraging `yup` for validation: https://www.npmjs.com/package/yup
- If needed, can define and add more detail about the custom validation rules required for each field.
- There will be some macros that we don’t to necessarily do right now but we’ll want to be mindful of, such as a Segment/Section
- Talk with Design about the pattern for the disabled state (based on not being connected)
- Drop in a rule into a config file that’s “disable if not connected”
- Build out the app and make it work and then we’ll revisit what we can optimize and refactor Form Builder and Transaction Builder features
  - Can we do this before Core UI? Perhaps we could get 90% there – create the packages -> refactor Summoner -> use in Core UI

jordanlesich:
As an add-on to this, I'm going add the internal display logic each form segment has.

PR #260, opened 2022-06-08
Summoner/validation form logic

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/60

## Changes

- Builds in internal logic for basic form validation
- Utilises react-hook-form's native validation system (was able to remove my own hacky system). 
- Builds form segment logic (changing UI based on formvalues)
- Handles field data modifiers (ex. parsing address, shares, loot from string and sending that data to formValues)
- Handles complex validations that can read post-modification data 
- Time Picker Input Select
- Many Typeguard and validation utils. 

Note: 

Did not end up using yup. While react-hook-form does have some decent integrations + docs on yup, it seems that yup likes a validation scheme at the top level of the form. The potential challenge I wanted to avoid was dealing with top-level validations, trying to time it with RHF validation timing and the custom modifiers we use. 

In the end, I decided to deal with the tedium of making some typecheck fns (or copying them over from V2) over dealing integration issues. I'm still open to reconsidering yup, but it would need to do the following: 

- validate at the field level and allow for components to have their own default validations/
- Handle custom value 
- Somehow work easily with the formBuilder API (would we pass in the schema as props?) when it gets built
- Allow for custom error messages (I'm sure it does that) 
- Allow for custom validations (I'm sure it does that as well; just making sure)

## Packages Added

"@gnosis.pm/safe-contracts" (will need for TX)

## Checks

- [x] App runs locally


apps/summon-app/src/layouts/ShamanSegment.tsx
```diff
@@ -1,22 +1,54 @@
...
       title="Starting Shamans"
       description="Shamans are very powerful as they can have administrative control over voting and non-voting stakes. Be very careful adding shamans. "
       formArea={
         <TextAreaSection>
           <TemporaryLink className="link">How to add a Shaman</TemporaryLink>
-          <ParSm className="number-display">0 Shamans</ParSm>
+          <ParSm className="number-display">{amtShamans} Shamans</ParSm>
           <WrappedTextArea
             label="Addresses & Permissions"
             placeholder="0x00000000000000000000000000 3"
```
skuhlmann, line 43: ZERO_ADDRESS might make a nice reusable constant.

apps/summon-app/src/utils/common.ts
```diff
@@ -0,0 +1,106 @@
...
+    'Shares are required and must be a number. Check formatting rules in tooltip above',
+  ADDRESS_ERR:
+    'Member addresses are required and must be valid Ethereum Addresses. Check formatting rules in tooltip above',
+  SHAMAN_ADDR_ERR:
+    'Shaman addresses are required and must valid ethereum addresses',
+  PERMISSION_ERR: 'Shaman permissions are required and must be a number',
+};
+
+////////////////////Members Segment////////////////////////
+export const validateMemberData = (
```
skuhlmann, line 19: :+1:

apps/summon-app/src/utils/summonModel.ts
```diff
@@ -0,0 +1,68 @@
...
+// );
+// const METADATA_CONFIG = {
+//   CONTRACT: LOCAL_ABI.BAAL,
+//   ACTION: 'executeAsBaal',
+//   ARGS: {
+//     TO: CONTRACT_ADDRESSES.POSTER['0x4'],
+//     VALUE: 0,
+//     DATA: METADATA,
+//   },
+// };
```
alexkeating, line 67: Should we remove these comments?

libs/ui/src/components/atoms/Button/Button.tsx
```diff
@@ -17,6 +17,7 @@ export type ButtonProps = {
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
   IconLeft?: IconType;
   IconRight?: IconType;
+  type?: 'button' | 'submit' | 'reset' | undefined;
```
alexkeating, line 20: Isn't undefined handled by the ?

alexkeating (approved):

apps/summon-app/src/utils/summonModel.ts
jordanlesich: It's there for when I put in the TX

apps/summon-app/src/layouts/ShamanSegment.tsx
jordanlesich: I'll add to commons

Issue #163, opened 2022-05-17
Create Summon Transaction


skuhlmann:

jordanlesich:
Going to add a TimePicker component to this.

PR #263, opened 2022-06-13
Summoner/tx

jordanlesich:
## GitHub Issue
https://github.com/HausDAO/daohaus-monorepo/issues/163
## Changes

- Builds a summon TX
- Checks types post-validation

Note: 
I use a lot of typecasting. Would be awesome if I could have a sync review to go over how to improve this. 

## Packages Added

NA

## Checks

- [x] App runs locally


skuhlmann (approved):
The tx logic looks good - similar to rage app with the new factory reqs incorporated.

apps/summon-app/src/layouts/SummonerForm.tsx
```diff
@@ -36,9 +40,12 @@ const Main = styled.main`
 `;
 
 export const SummonerForm = () => {
+  const { provider, chainId } = useHausConnect();
   const methods = useForm({ mode: 'onTouched' });
   const handleFormSubmit = async (formValues: Record<string, unknown>) => {
```
alexkeating, line 45: Maybe the types can be something like this? https://github.com/raid-guild/headline/blob/6b6ac5b00dbacff0253418c43abf7af0858f49d0/src/pages/PublishPage.tsx#L337

SubmitHandler<FieldValues> as the return type might resolve some of the issues lower down in the function call tree

apps/summon-app/src/utils/summonTx.ts
```diff
@@ -0,0 +1,265 @@
...
+
+  const memberData = formValues[FORM_KEYS.MEMBERS] as Record<string, unknown>;
+
+  if (
+    !isArray(memberData?.memberAddresses) ||
+    memberData.memberAddresses.some((addr) => !isString(addr)) ||
+    !isArray(memberData?.memberShares) ||
+    memberData.memberShares.some((shares) => !isNumberish(shares))
+  ) {
+    console.log('ERROR: Form Values', formValues);
```
alexkeating, line 103: There are a couple of these that might be good to remove

libs/common-utilities/src/constants/contracts.ts
```diff
@@ -0,0 +1,19 @@
...
+  },
+  V3_FACTORY: {
+    '0x5': '0x1b94221EB2bC8dc9F16660EA5be9dcd92b0ae862',
+  },
+  GNOSIS_MULTISEND: {
+    '0x5': '0xA238CBeb142c10Ef7Ad8442C6D1f9E89e07e7761',
+  },
+  POSTER: {
+    '0x5': '0x000000000000cd17345801aa8147b8d3950260ff',
+  },
```
alexkeating, line 18: We should probably change the keychain type to be be a network object rather than a string

apps/summon-app/src/utils/summonTx.ts
```diff
@@ -0,0 +1,265 @@
...
+    V3_LOOT_SINGLETON,
+    V3_SHARE_SINGLETON,
+    GNOSIS_MULTISEND,
+    POSTER,
+  ];
+
+  if (v3Contracts.every((contract) => contract[chainId])) {
+    // REVIEW. Before this conditional statement, each contract
+    // was string | undefined. After the conditional statement it can
+    // only be string. So I'm typecasting here.
```
alexkeating, line 188: There is probably some disconnect here with typescript. If we are looping through every item it might make sense to combine into a netowork object ad then just access the keys to avoid the typecasting

alexkeating (commented):
Potentially alot of the type guards can be avoided by changing Record<string, unknown>. Hopefully, my suggestion helps

Issue #57, opened 2022-04-18
Integrate Dao-Data Polling in Summoner App


jonathanprozzi:
## Overview

We've done a Polling Rage, but this is a different use case. This also represents  the work around a deeper dive into polling and combining it with other features such as caching. 

https://github.com/HausDAO/daohaus-monorepo/issues/48

## Details

These aren't specifically broken into any particular TODOs at this point, but representing the general work areas for this.

- Integrate polling with caching 
- Integrate polling for **user initiated transactions** or by **transactions in the cache**
- SDK Considerations:
  - API where someone can choose a continuous poll similar to the Rage or user-initiated poll (such as what's implemented in v2)
    - Need to integrate the state – should be unopinionated
    - Possibly a callback, but what function would be called?

## Deliverables

- Implementation of Polling based on the specific Summoner use cases outlined above
- Consider SDK implications and whether or not anything should be added to the SDK
- PR connected to this issue

## Related Issues

- #59 -- This could be an opportunity to pair on these since they're related

PR #264, opened 2022-06-13
Feature/tx query

skuhlmann:
## GitHub Issue

related to this - but it doesn't close this issue:
[Add a link to the GitHub issue.](https://github.com/HausDAO/daohaus-monorepo/issues/57)

## Changes

- new endpoint to query on tx

## Packages Added

n/a

## Checks

- [ ] App runs locally


libs/dao-data/src/Query.ts
```diff
@@ -317,7 +320,7 @@ export default class Query {
     }
   }
 
-  public async findLatestTransaction({
+  public async findLatestTransactionByDao({
```
alexkeating, line 323: Should we allow clients to pass in the where clause?

skuhlmann: Yeah that's fair. this was a super scoped function for post proposal tx query... but was anticipating a usecase we might not need.

going to convert to a more typical list query and we can pass the where, sort order and limit like this is doing.

skuhlmann: @alexkeating  new commit has this. much cleaner and more flexible. thanks!

libs/dao-data/src/Query.ts
```diff
@@ -317,13 +358,13 @@ export default class Query {
     }
   }
 
-  public async findLatestTransaction({
+  public async findTransaction({
```
jordanlesich, line 361: This looks like it works for me.

jordanlesich (approved):
Not a whole lot to report here that already hasn't been said/fixed already. I'd be happy merging this in

Issue #237, opened 2022-06-03
Create Toast Component


jonathanprozzi:
## Overview

Create a `<Toast/>` component that supports several different variants (success and error, for example) can be used throughout our apps.

![Image](https://user-images.githubusercontent.com/9438776/171768868-b2adefed-0f3c-4aeb-b561-4f7dfec5818e.png)

## Deliverables

- [ ] A `<Toast/>` component added to the Component Library
- [ ] Stories added to the Storybook for the Toast variants

## Details

- We have success and error variants, and examples of each that support the following:
  - Icon (such as success and error icons in the top left)
  - Left link
  - Right link
  - Left and right links

## Related Issues

- #230 

## Resources

- [Pattern Library Figma Link](https://www.figma.com/file/R1a1bwODnzCHHHw0EJALAU/DAOhaus-v3-Pattern-Library-(In-Progress)?node-id=797%3A92905)

earth2travis:
Any updates on this @0xRowdy?

earth2travis:
@7helles we may need to add a warning state for toasts in the Pattern Library

PR #301, opened 2022-06-24
Feature/toast component

0xRowdy:
## GitHub Issue

[GitHub issue](https://github.com/HausDAO/daohaus-monorepo/issues/237)

## Changes

Created Toast Component
Created Card Component
Updated Link Component

## Packages Added

Radix Toast component
yes need to rerun `yarn install` to run locally

## Checks

- [x] App runs locally


libs/ui/src/components/molecules/Toast/Toast.styles.ts
```diff
@@ -0,0 +1,97 @@
+import * as ToastPrimitive from '@radix-ui/react-toast';
+import styled, { keyframes } from 'styled-components';
+
+import { font } from '../../../theme/global/font';
```
jordanlesich, line 4: Will need to fix these warnings

libs/ui/src/components/molecules/Toast/Toast.tsx
```diff
@@ -0,0 +1,142 @@
...
+          </ToastHeaderContainer>
+          {toastLinks && <ToastLinks {...toastLinks} />}
+        </Card>
+      </ToastRoot>
+      <ToastViewport label={label} hotkey={hotkey} />
+    </ToastProvider>
+  );
+};
+
+// If Toast contains links
```
jordanlesich, line 117: My initial thinking was that the toast could be more generic, allowing for possible any sort of content to placed within it. Seeing this, I'm thinking the choice of dialing it in a bit more is better and definitely more in line with that we're doing with the component library. Well done.

jordanlesich (approved):
The code looks clean and organized. I like the structure and I think it'll be easy to use. 

I'd say just fix the warnings and we're good. I'll test it out in storybook later today.

Issue #230, opened 2022-06-03
Create Toast Hook


jonathanprozzi:
## Overview

Create a hook to support use of Toasts throughout our app. We may want to leverage a similar `OverlayContext` pattern used throughout v2.

## Deliverables

- [ ] A `useToast` hook added to support our Toast components
- [ ] (optional) `OverlayContext` pattern (or at least testing to see if our implementation would support this approach) -- defer to you on this one @jordanlesich!

## Details

- Should we look to refactor/reuse patterns such as the in the `OverlayContext` in v2? (https://github.com/HausDAO/daohaus-app/blob/develop/src/contexts/OverlayContext.jsx)
  - This version leverages the `useToast` hook from [Chakra](https://github.com/chakra-ui/chakra-ui/blob/main/packages/toast/src/use-toast.tsx), so we may be able to look to this for inspiration for this hook

## Related Issues

- #237 

## Resources

- [Chakra UI useToast Code](https://github.com/chakra-ui/chakra-ui/blob/main/packages/toast/src/use-toast.tsx)

jonathanprozzi:
@jordanlesich -- i added some additional notes here, but i know you also have more context for this than i do!

jordanlesich:
@jonathanprozzi Notes are good. Going to jam this in HausThemeContext I think.

PR #305, opened 2022-06-26
Feature/toast hook

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/230

## Changes

See issue description. 

Current limitations: 
- Does not handle multiple toasts atm. We need to see how animation shakes out first. 

## Packages Added

NA

## Checks

- [x] App runs locally

### DIFF
--- a/libs/ui/src/components/molecules/Toast/Toast.tsx
+++ b/libs/ui/src/components/molecules/Toast/Toast.tsx
@@ -25,7 +25,7 @@ import {
   CloseIcon,
 } from './Toast.styles';
 
-type CustomToastProps = ToastProps &
+export type CustomToastProps = ToastProps &
   ToastProviderProps &
   ToastViewportProps & {
     title: string;
@@ -38,7 +38,7 @@ type CustomToastProps = ToastProps &
     toastLinks?: ToastLinksProps;
   };
 
-type ToastLinksProps = {
+export type ToastLinksProps = {
   actionAltText?: string;
   leftLink?: {
     path: string;
--- a/libs/ui/src/components/molecules/Toast/Toast.tsx
+++ b/libs/ui/src/components/molecules/Toast/Toast.tsx
@@ -1,9 +1,4 @@
 import {
-  ToastProps,
-  ToastProviderProps,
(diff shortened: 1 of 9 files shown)


libs/ui/src/components/molecules/Toast/Toast.tsx
```diff
@@ -38,7 +38,7 @@ type CustomToastProps = ToastProps &
     toastLinks?: ToastLinksProps;
   };
 
-type ToastLinksProps = {
+export type ToastLinksProps = {
```
0xRowdy, line 41:
```suggestion
type ToastLinksProps = {
```
Don't think we're using ToastLinkProps anywhere outside this file.

libs/ui/src/hooks/useToast.tsx
```diff
@@ -0,0 +1,56 @@
...
+    title: string;
+    description?: string;
+  }): void => {
+    setToast({
+      title,
+      description,
+      error: true,
+      iconType: 'error',
+    });
+  };
```
0xRowdy, line 20:
```suggestion
  };
  
```
Line break?

libs/ui/src/hooks/useToast.tsx
```diff
@@ -0,0 +1,56 @@
...
+    title: string;
+    description?: string;
+  }): void => {
+    setToast({
+      title,
+      description,
+      success: true,
+      iconType: 'success',
+    });
+  };
```
0xRowdy, line 34:
```suggestion
  };
  
```

0xRowdy (approved):

libs/ui/src/components/molecules/Toast/Toast.tsx
jordanlesich: I'm using it in the context now. I can refactor it somewhere higher up.

0xRowdy: It looked like CustomToastProps was used in context. ToastLinkPros is a part of CustomToastProps.

Issue #244, opened 2022-06-07
Create DAO Card Component


jonathanprozzi:
## Overview

Create the DAOs Component card for use in the Grid view. These take in the same data used in the _DAO Table_ view and display the DAO's info as a card.

![Image](https://user-images.githubusercontent.com/9438776/175979205-409337c4-7ae3-43f6-a3b7-91a568b7c1c7.png)

## User Stories

- As a connected user viewing the _Grid view_, each DAO Card should display information about a DAO I'm in.
- As a connected user who is a delegate, the DAO Card for the DAO I'm a delegate in should display a tag showing this.

## Details

- DAO logo/icon (proposal count is a stretch goal -- can add later in a separate ticket)
- If user is the Delegate for the DAO, display a _Delegate_ tag in the top right
- Each card should display the following as text:
  -  DAO name
  - Member count
  - Treasury/vault value
  - Proposal count
  - Voting power

- Each card should display the following as tags:
  - Chain
  - DAO type (ie "Moloch V2")

- Link to the DAO on DAOhaus v2

## Related Issues

- #243 is the grid view that uses these cards

earth2travis:
How does this relate to https://github.com/HausDAO/daohaus-monorepo/issues/258

skuhlmann:
> How does this relate to #258

Oh we missed that one Ven had added. #258 can be the ticket of truth.

jonathanprozzi:
Closed by PR https://github.com/HausDAO/daohaus-monorepo/pull/314

PR #314, opened 2022-06-28
Dao card

jordanlesich:
## GitHub Issue
https://github.com/HausDAO/daohaus-monorepo/issues/244
## Changes

See Issue

Screenshot from 2022-06-28 00-31-51

Screenshot from 2022-06-28 00-31-25

If we pass in a link to the avatar it should render something other than this ugly box. Perhaps we can think of an SVG to add as fallback? 

Not doing the separate colors for tags until we refactor to component library. Should make a ticket for this (this one should be included in this week's push; other components in here could wait probably)

## Packages Added

NA

## Checks

- [x] App runs locally


jordanlesich:
Lots of comments in here for Component library and design.

jordanlesich:
Two tickets we should make for this push of Hub are the Tag and tag color system and unit abbreviation utilities.   

A tag component should take a color prop (Enum) that changes text and bg color based on that. This gets a little bit tricky with the way Radix is with their color object naming system (but still pretty easy if we limit colors). Then we pass it the text as children. Worth figuring out if we want tags that also conform to the theme object as well. 

Unit manipulation utils. I had to build these in V2. They just add things like 'k' or 'e' to number amts that are over a certain size. I used a lib. Forget which one. We need these for a lot of things in the hub app.

0xRowdy (approved):

PR #319, opened 2022-06-29
Hub/disconnected

jordanlesich:
## GitHub Issue

https://github.com/orgs/HausDAO/projects/3/views/3

## Changes

See issue. 

## Packages Added

NA

## Checks

- [x] App runs locally


apps/hub-app/src/components/DaoCards.tsx
```diff
@@ -11,7 +12,8 @@ const CardListBox = styled.div`
   flex-wrap: wrap;
   column-gap: 6rem;
   row-gap: 3rem;
-  .dao-card {
+  @media ${widthQuery.sm} {
```
alexkeating, line 15: Oh this breaks away from the other media queries with min-width. The idea being that media queries encapsulate styles for views larger than mobile

jordanlesich: Switched queries here and in DAOCards to mobile first.

Issue #36, opened 2022-04-06
Deploy Sage Storybook


jonathanprozzi:
## What

Deploy the Sage Storybook to either [Chromatic](https://www.chromatic.com/), GitHub Pages, or s3. Chromatic is a free cloud host from the Storybook maintainers. 

We can start with Chromatic, GitHub Pages, or s3 and then maybe look at using Skynet or a decentralized host like Fleek.

We can likely deploy only the Sage version of our Storybook (instead both Rage and Sage) at a certain point, although we may want to keep both so that we have somewhat of a _staging_ and _prod_ type setup for our component library.

## TODOs

- [x] Determine deploy target 
- [x] Add deploy command to root `package.json` in the Sage monorepo
- [ ] Add deploy workflow to GitHub Actions

## Deliverables

Storybook build commands added to the root's `package.json` and a hosted Storybook instance with CI/CD based on the `haus-components` package.

## Resources

- https://storybook.js.org/tutorials/intro-to-storybook/react/en/deploy/

scottrepreneur:
Check out the deploy workflow being used by the RG system.
https://github.com/raid-guild/design-system/blob/main/.github/workflows/publish.yml

jonathanprozzi:
Making progress on this -- had to make a few adjustments for our workflow and also had to think through a few things with regards to our monorepo @earth2travis

jonathanprozzi:
We'll be working on this on Weds.

jonathanprozzi:
This is resolved by #379

earth2travis:
Blocked by https://github.com/HausDAO/daohaus-monorepo/issues/398

earth2travis:
@jonathanprozzi looks like https://github.com/HausDAO/daohaus-monorepo/issues/398 is ready for review

PR #379, opened 2022-07-06
feature/deploy-storybook

jonathanprozzi:
## GitHub Issue

#36 

## Changes

Added a `deployStaging` command in the `project.json` in the ui library.

This is successfully deploying, but we discovered an error when serving the built Storybook that we can address in another issue. I'll create a card for this.

## Packages Added

n/a

## Checks

- [x] App runs locally


skuhlmann:
@jonathanprozzi This looks good - but curious if we are going to merge this before the deployed storybook is working. Did that issue get rexolved?

jonathanprozzi:
> @jonathanprozzi This looks good - but curious if we are going to merge this before the deployed storybook is working. Did that issue get rexolved?

Good call! it's still broken, so we should definitely not merge this in until that's fixed. I'll take a deeper look at it

alexkeating:
@jonathanprozzi @jordanlesich @skuhlmann This should be good now. The issue was the NODE_ENV needs to be set to production on build storybook

PR #424, opened 2022-07-12
Example to add declaration file

alexkeating:


jordanlesich:
This is just an example, right? We're not merging anything through?

Issue #431, opened 2022-07-13
Sensible defaults for lib package.json


brossetti1:
# Issue

I noticed our npm packages do no currently display certain details that are basic and useful for display in npm and github.

# Solution

add some sensible defaults. if this is not warranted yet, feel free to close the pr and issue

PR #432, opened 2022-07-13
adding sensible defaults to package.json for npm to read

brossetti1:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/431

## Changes

I noticed our npm packages listings do not have much config details being read for our libs. Adding some basic details package.json so they display in npm and github based on reading package.json

 if this is not warranted yet, feel free to close the pr and issue

## Packages Added

N/A

## Checks

- [X] App runs locally

# Details

Current dao haus lib packages:

Screen Shot 2022-07-13 at 2 46 05 PM

React.js npm (notice Licnese field, issues, homepage, etc):

Screen Shot 2022-07-13 at 3 05 21 PM

### DIFF
--- a/libs/abi-utilities/package.json
+++ b/libs/abi-utilities/package.json
@@ -1,5 +1,18 @@
 {
   "name": "@daohaus/abi-utilities",
   "version": "0.0.2",
-  "type": "commonjs"
+  "type": "commonjs",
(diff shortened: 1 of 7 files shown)


jonathanprozzi (approved):
I think this is a great idea -- the only real reason we didn't as far as I know is due to a time crunch when we scaffolded for release

Great addition

PR #433, opened 2022-07-13
Core UI/scaffold

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/422
## Changes

- Scaffolds new app
- Adds in stupid human format dts
- Changes main to use React 17 API
- Fixes window.global issue

## Packages Added

## Checks

- [x] App runs locally


jordanlesich:
Getting some issues around the prebuilt app tests. Deleting the tests fixes it locally. Let's see if it works here.

apps/core-app/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="/src/favicon.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Vite App</title>
```
jonathanprozzi, line 7: We may as well change this to something else and add the DH favicon too -- What would we call Core UI for a user-facing name?

Not a priority, just noting

apps/core-app/src/polyfills.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Polyfill stable language features. These imports will be optimized by `@babel/preset-env`.
+ *
+ * See: https://github.com/zloirock/core-js#babel
+ */
+import 'core-js/stable';
+import 'regenerator-runtime/runtime';
```
jonathanprozzi, line 7: ah interesting. i ran into this in another vite app. was causing unexpected issues so good catch!

jordanlesich: I think it does this by default.

jonathanprozzi (approved):
Looks good! Question around what we're calling this as user facing though -- Will this just be known as 'DAOhaus v3' to users while we continue to call it Core UI/Core App internally?

skuhlmann:
> Looks good! Question around what we're calling this as user facing though -- Will this just be known as 'DAOhaus v3' to users while we continue to call it Core UI/Core App internally?

Good question. I think getting some content strategy from des9gn would be good on that.

Lately I've been thinking of it as the dao admin tool. admin.daohaus.club feels like a cool framing to me.

jordanlesich:
Updated the favicon and title. Went with 'DAOhaus Core" for now. 

Imo, V3 has sort of become this moniker for all the apps and tools + MolochV3.  However, would be good to hear from design and see what would be most useful for users.

jordanlesich:
Oh, admin seems really intuitive. I like that.

PR #443, opened 2022-07-14
Bug/example

jordanlesich:
## Issue

This is an example of code that has a typescript bug, but passes build and lint in the monorepo, and in CI 

This branch passes when I run the following commands locally: 
- nx lint ui
- nx build ui
- nx lint summon-app
- nx build summon-app

The actual bug is a component that is sending props to component that no longer recieves those props. Very similar error to the last one. 

If I go to that component in VSCode, the error is highlighted: 

Screenshot from 2022-07-14 14-44-49

It also passes build on CI, but that's because no changes were targeted. My thinking is that it's because these changes were already merged to develop. However, that should mean that it was targeted on the branch where these changes were initially made. 

Screenshot from 2022-07-14 14-54-50

Saving this branch to help find the bug and fix it. Let me know if you have any questions.


alexkeating:
Where is the bug in this pr? Is it in the diff? Or is it in one of the components? And yes if the changes are in develop then affected will not run

alexkeating:
I found it in your other pr

Issue #238, opened 2022-06-03
Add Transaction Hash and Timestamp Data to Subgraph Entities


skuhlmann:
# Description

Core UI will have links to transactions on block explorers for many actions/activities. And add to the SDK queries where necessary.

Add thee hash from `event.transaction.hash` and `event.block.timestamp` for actions in the following event handlers.

and some from `event.block.timestamp` if the createdAt field doesn't already tell us that.

and some for the address the is performing the tx - event.transaction.from

- [x] handleSummonBaal

entity: Dao
new fields: summonTxHash
createdBy: createdBy

- [x] handleSubmitProposal

entity: Proposal
new field: proposalSubmissionTxHash

- [x] handleSponsorProposal

entity: Proposal
new field: proposalSponsorTxHash
timestamp field: proposalSponsorTxAt 

- [x] handleProcessProposal

entity: Proposal
new field: proposalProcessTxHash
timestamp field: proposalProcessTxAt 
createdBy: processedBy

- [x] handleCancelProposal

entity: Proposal
new field: proposalCancelTxHash
timestamp field: proposalCancelledTxAt 
createdBy: cancelledBy

- [x]  handleSubmitVote

entity: Vote
new field: voteSubmisssionTxHash

- [x] handleRageQuit

entity: RageQuit
new field: rageQuitTxHash

- [x] handleSharesTransfer & handleLootTransfer

entity: Member
new field: memberCreatedTxHash

- this one will be on the new Member creation in mintLoot/mintShares

- [x] handleDelegateChanged

entity: Member
new field: lastDelegateUpdateTxHash

- [x] update sdk queries

skuhlmann:
then add these into the queries in the sdk

skuhlmann:
Once we deploy these to the hosted graph we'll need to regenerate the types for the new subgraph queries. Once you make the changes from my review we can deploy it.

brossetti1:
@skuhlmann updated

PR #468, opened 2022-07-18
adding timestamp and address fields to many events

brossetti1:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/238

## Changes

adding several timestamp and address details to Dao, Proposals, Votes, and Member

includes build files

## Packages Added

N/A

## Checks

- [X] App runs locally


brossetti1:
@skuhlmann i noticed Dao already had `transactionHashSummon` so skipped `summonTxHash`

wondering if we need to update fields in:

transformProposal : https://github.com/HausDAO/daohaus-monorepo/blob/develop/libs/dao-data/src/utils/transformers.ts#L15
transformMembershipList: https://github.com/HausDAO/daohaus-monorepo/blob/develop/libs/dao-data/src/utils/transformers.ts#L56

apps/v3-subgraph/schema.graphql
```diff
@@ -103,10 +107,26 @@ type Proposal @entity {
   sponsored: Boolean!
   "address that sponsored the proposal"
   sponsor: Bytes!
+  "transaction hash of the proposal sponsor"
+  proposalSponsorTxHash: Bytes!
```
skuhlmann, line 111: This should probably be nullable as we won't have data for that field when saving the proposal upon submission.

apps/v3-subgraph/schema.graphql
```diff
@@ -103,10 +107,26 @@ type Proposal @entity {
...
+  "unix timestamp of when the proposal was sponsored"
+  proposalSponsorTxAt: BigInt!
```
skuhlmann, line 113: nullable

apps/v3-subgraph/schema.graphql
```diff
@@ -103,10 +107,26 @@ type Proposal @entity {
...
   "indicates if the proposal is cancelled"
   cancelled: Boolean!
+  "transaction hash of the cancelled proposal"
+  proposalCancelTxHash: Bytes!
```
skuhlmann, line 117: same with these - all the proposal states that might not happen or don't happen until later in the proposal lifecycle will need to be nullable.

sponsor
cancelled
processed

apps/v3-subgraph/schema.graphql
```diff
@@ -194,6 +218,8 @@ type Member @entity {
   delegatingTo: Bytes!
   "subgraph id of member the address is delegating to"
   delegatingToMember: Member
+  "the transaction hash when the delegate was last updated"
+  lastDelegateUpdateTxHash: Bytes!
```
skuhlmann, line 222: nullable

skuhlmann (changes requested):

skuhlmann (approved):

Issue #406, opened 2022-07-12
Break Down Form Builder tasks


jordanlesich:

earth2travis:
See https://github.com/HausDAO/daohaus-monorepo/issues/292

PR #499, opened 2022-07-21
Feature/form builder

jordanlesich:
## GitHub Issue
https://github.com/HausDAO/daohaus-monorepo/issues/406
## Changes

See issue
## Packages Added

NA

## Checks

- [x] App runs locally


libs/form-builder/src/components/CoreFieldLookup.ts
```diff
@@ -0,0 +1,21 @@
...
+  WrappedInput,
+  WrappedInputSelect,
+  WrappedRadio,
+  WrappedSelect,
+  WrappedSwitch,
+  WrappedTextArea,
+} from '@daohaus/ui';
+import { SplitColumnLayout } from './SplitRender';
+
+export const CoreFieldLookup = {
```
jonathanprozzi, line 12: this is very cool -- allows for us to extend when/if we need to add and map more inputs for the builder

libs/form-builder/src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,69 @@
+import { FormProvider, useForm } from 'react-hook-form';
+
+import { FormLayout } from '@daohaus/ui';
```
jonathanprozzi, line 3: One thing I'm wondering for down the road -- I wonder if we'd want to consider doing a styleless version of this similar to react-table

libs/form-builder/src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,69 @@
...
+import { FormBuilderFactory } from './FormBuilderFactory';
+import { FormLego } from '../types/legoTypes';
+
+import { Logger } from './Logger';
+import { FormFooter } from './formFooter';
+import { useState } from 'react';
+import { isValidNetwork } from '@daohaus/common-utilities';
+import { useHausConnect } from '@daohaus/daohaus-connect-feature';
+
+type FormBuilderProps = FormLego & {
```
jonathanprozzi, line 14: Nice!

libs/form-builder/src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,69 @@
...
+    setIsSubmitting(false);
+  };
+  return (
+    <FormLayout title={title} subtitle={subtitle} description={description}>
+      <FormProvider {...methods}>
+        <form
+          onSubmit={methods.handleSubmit(handleTopLevelSubmit)}
+          className="builder-inner-form"
+        >
+          <Logger log={log} />
```
jonathanprozzi, line 56: Since log is a boolean do we need to handle if `log={false}` ? If not do you think we'd want to bake this in w/o allowing folks to choose?

libs/form-builder/src/components/Logger.tsx
```diff
@@ -0,0 +1,13 @@
+import { useEffect } from 'react';
+import { useWatch } from 'react-hook-form';
+
+export const Logger = ({ log }: { log: boolean }) => {
+  const formValues = useWatch();
+  useEffect(() => {
+    if (log && formValues) {
```
jonathanprozzi, line 7: I see we're handling if log is false here, but wonder if we'd need to handle it in the FormBuilder where log is called. Probably not a big deal as it would be handled at this component but wondering if it would save us an unnecessary render of Logger if log={false}

libs/form-builder/src/components/StandardFields.ts
```diff
@@ -0,0 +1,88 @@
...
+    type: 'input',
+    required: true,
+    info: "This is a description of the field's purpose.",
+  },
+  Switch: {
+    id: 'id2',
+    label: 'Label',
+    type: 'switch',
+    required: true,
+    switches: [{ fieldLabel: 'shit' }],
```
jonathanprozzi, line 16: 👀

libs/form-builder/src/components/formFooter.tsx
```diff
@@ -0,0 +1,11 @@
+import { Button } from '@daohaus/ui';
+import React from 'react';
+
+export const FormFooter = ({ submitDisabled }: { submitDisabled: boolean }) => {
+  /*Form Alert Component goes here*/
```
jonathanprozzi, line 5: What is the Form Alert component? I'm sure that this is something I should already know 😅

libs/ui/src/components/molecules/WrappedInput/WrappedInput.tsx
```diff
@@ -9,6 +9,8 @@ export const WrappedInput = (props: Field) => {
   const registration = registerOptions
     ? register(id, registerOptions)
     : register(id);
+
+  console.log('id', id);
```
jonathanprozzi, line 13: We can probably remove this unless it's helpful for debugging while this is still under development/iteration

jonathanprozzi (approved):
This is really cool @jordanlesich -- great work! The patterns look great. I haven't had the chance to fully dig in and test it yet but I want to approve it to get the ball rolling on integrating this into Core UI to help streamline our development! Stoked to play with this.

libs/form-builder/src/components/FormBuilder.tsx
jordanlesich: Yeah, that's a good idea.

libs/form-builder/src/components/FormBuilder.tsx
```diff
@@ -0,0 +1,69 @@
...
+  description,
+  fields,
+  onSubmit,
+  log,
+}: FormBuilderProps) => {
+  const methods = useForm();
+  const {
+    formState: { isValid },
+  } = methods;
+  const { chainId } = useHausConnect();
```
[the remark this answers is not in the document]
jordanlesich: Yeah, was thinking builders would be super high-level for now. But maybe I could make this props? 

It would be awesome if we could find a way to use connect or anything else. 

I'd be down if we made a ticket to remove these deps down the road.

libs/form-builder/src/components/Logger.tsx
jordanlesich: Probably cleaner. I can clean this up more.

libs/form-builder/src/components/formFooter.tsx
jordanlesich: It's that alert that goes near the submit button. It lets you know that you're not on the right network or whatever.

libs/ui/src/components/molecules/WrappedInput/WrappedInput.tsx
jordanlesich: Whoops.

libs/form-builder/src/components/SplitRender.tsx
```diff
@@ -0,0 +1,26 @@
...
+  rows: { rowId: string; left: FieldLego; right: FieldLego }[];
+};
+
+export const SplitColumnLayout = ({ rows, ...props }: SplitColumnProps) => {
+  return (
+    <SplitColumn
+      rows={rows.map(({ left, right, rowId }) => {
+        return {
+          rowId,
+          left: <FormBuilderFactory {...left} {...props} spacing={false} />,
```
0xRowdy, line 20: Are there potential concerns around implementing the factory in two places? I don't have knowledge of state management.

0xRowdy (approved):

libs/form-builder/src/components/SplitRender.tsx
jordanlesich: Let's talk through the this. From my understanding, since the factory is stateless it should be fine. Recursive rendering is a pattern is a pattern I'm hoping to user throughout the builders. So it would be good to highlight any challenges we foresee,

Issue #194, opened 2022-05-24
Set up DAO Producer Job


alexkeating:
## Overview

Once our Kubernetes Cluster is set up we can create jobs that run. This job will be used in our Hub App for DAO processing. 

## Details

- Create a job that pushes DAOs to queue to be processed
  - Query subgraphs for each DAO in the queue  
  - Drops in the DAO address, safe address fields, network (needed to query for token balances) 

## Deliverables

- DAO Producer Job set up in our infrastructure that supports the Hub App data processing needs

## Related Issues

- #195 
- #172 
- #196

earth2travis:
Requires Helm

alexkeating:
Waiting on this pr to get merged to resolve the esm issues https://github.com/nrwl/nx/pull/10414

PR #503, opened 2022-07-25
Feature/producer new

alexkeating:
## GitHub Issue

#194 

## Changes

- Add producer job with corresponding helm charts
- Migrate nx to latest verison

## Packages Added

## Checks

- [ ] App runs locally


apps/dao-producer-job/src/main.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const main = async () => {
+  console.log('Start processing daos');
+  const producerQueue = new Queue('dao-producer', REDIS_HOST);
+  try {
+    const haus = Haus.create();
+    for (const networkId of networks) {
+      console.log(`Starting network ${networkId.toString()}`);
+      const resp = await haus.query.listDaos({ networkId });
```
skuhlmann, line 31: This will work well for now - but we'll need to deal with pagination at some point.

basicall the resp will give you a new prop called nextPaging that you'll pass on a subsequent listDaos call until nextPaging is undefined.

```
haus.listDaos({netoworkId, paging: resp.nextPaging})
```

alexkeating: Good call

alexkeating: Updated pagination

apps/dao-producer-job/src/main.ts
```diff
@@ -0,0 +1,59 @@
...
+      do {
+        const nextPage = resp?.nextPaging ? { paging: resp.nextPaging } : {};
+        resp = await haus.query.listDaos({ networkId, ...nextPage });
+        for (const dao of resp.items) {
+          console.log(`Pushing up dao ${dao.id}`);
+          producerQueue.add({
+            address: dao.id,
+            safeAddress: dao.safeAddress,
+            network: networkId,
+          });
```
brossetti1, line 41: so it looks like we are mainly setting up queuing a dao, we have not implemented any queue logic currently?

brossetti1 (approved):
looks good, left one question. im wondeirng if alot of these files (minus the nx migration files) are generated by helm?

if so, is there a reference to the generator commands for adding jobs?

alexkeating:
> looks good, left one question. im wondeirng if alot of these files (minus the nx migration files) are generated by helm?
> 
> if so, is there a reference to the generator commands for adding jobs?

No most of the files were from the nx migration

apps/dao-producer-job/src/main.ts
alexkeating: Yes, Queue is simple and logic will be handled in the consumer

Issue #458, opened 2022-07-18
Baal: Update SummonBaal event to indicate with summon function was used


skuhlmann:
# Details

We need a way to flag if a baal was summoned from `function summonBaal` vs. `function summonBaalAndSafe`

They both use this event:

```
    event SummonBaal(
        address indexed baal,
        address indexed loot,
        address indexed shares,
        address safe
    );

```

Can we add a 5th param to indicate if the safe was summoned with the baal or if it already existed?

brossetti1:
@skuhlmann @dekanbro after todays discussion around the gnosis plugin - should this field be considered for responding with multiple places a Baal can be summoned?

gnosis app vs summonBaal vs summonBaalAndSafe?

might make the bool irrelavent.. could probably come later

skuhlmann:
> @skuhlmann @dekanbro after todays discussion around the gnosis plugin - should this field be considered for responding with multiple places a Baal can be summoned?
> 
> gnosis app vs summonBaal vs summonBaalAndSafe?
> 
> might make the bool irrelavent.. could probably come later

Well this field really just indicates that the safe existed before the dao was summoned - we can't exactly tell where the ball was summoned from. In the future I think we'll support summoning with an existing safe in our UI too. 

The only thing we can really differentiate here is whether the summonBaal or summonBaalAndSafe function was called, so bool seems good to me.

dekanbro:
Summoning Baal factories on network: Goerli
Account address: 0xC00e02e6539376465190B2cA4759Bff6FE5f5B81
Account balance: 0.29748152283185087 ETH
lootSingleton 0x0De84DCAc3B2d52581120059ee9723FDDecCB044
sharesSingleton 0x3109AeD0fD9777cEFb24dBa5eb5030987bd9E3F3
baalSingleton 0x69b442eb55714A0B144134AED015517394Ed1871
Transaction Hash: 0x62906ba23728bda0a1a0ffcca412371772448bac200497030462da615fc04598
**Factory Contract Address**: 0x0C5fd8AAdF995e11E5Ac1CD72139Ee4fd72cDeFC

PR #520, opened 2022-07-26
Fix/summon event order

skuhlmann:
## GitHub Issue

[issue](https://github.com/HausDAO/daohaus-monorepo/issues/458)

## Changes

- baal summoner mapping update

## Checks

- [ ] App runs locally


apps/v3-subgraph/subgraph.yaml
```diff
@@ -1,4 +1,4 @@
-specVersion: 0.0.4
+specVersion: 0.0.2
```
jonathanprozzi, line 1: Probably something I'm missing, but why drop the version?

jonathanprozzi (approved):

brossetti1 (approved):

apps/v3-subgraph/subgraph.yaml
skuhlmann: Good question, not sure why/how that is happening. Will check it out

Issue #266, opened 2022-06-14
Create Layout Component


oovg:
# Details

Component to support the one-column layouts seen in core ui:
- Vault/members/settings pages
https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6037%3A140410
https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=5359%3A118119

**Overview and member profile page**
https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=5522%3A113968
https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6037%3A140410

- [x] optional page title
- [x] optional 'actions' button area

#### Legacy notes

Create components to replace `<TemporaryLayout />`

Should be high level page layout components for `hero` and `body`

earth2travis:
@oovg could you sync up with @jordanlesich to determine if what was implemented is okay?

PR #579, opened 2022-08-01
Feature/266 single col layout

skuhlmann:
## GitHub Issue

[GitHub issue.](https://github.com/HausDAO/daohaus-monorepo/issues/266)

## Changes

- single column layout
- still supports title, subtitle, description
- can pass actions (buttons)
- can pass children and deal with their width on their own

Screenshot from 2022-08-01 14-58-27
Screenshot from 2022-08-01 14-58-18

## Checks

- [ ] App runs locally


libs/ui/src/components/layouts/SingleColumnLayout/SingleColumnLayout.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const sectionSubtitle = useMemo(() => {
+    if (!subtitle) return null;
+    if (typeof subtitle === 'string')
+      return <DataMd className="subtitle">{subtitle}</DataMd>;
+    return <div className="subtitle">{subtitle}</div>;
+  }, [subtitle]);
+
+  const sectionTitle = useMemo(() => {
+    if (!title) return null;
+    let titleNode = <div className="title">{title}</div>;
```
jordanlesich, line 35: returns instead of variable assigment (state)

libs/ui/src/components/layouts/SingleColumnLayout/SingleColumnLayout.tsx
```diff
@@ -0,0 +1,69 @@
...
+      titleNode = <H2 className="title">{title}</H2>;
+    }
+    if (!actions) return titleNode;
+    return (
+      <TitleContainerWithActions>
+        {titleNode}
+        {actions && (
+          <ActionButtonContainer>
+            {actions.map((action) => (
+              <div key={uuid()}>{action}</div>
```
jordanlesich, line 46: I'm moving away from uuid() keys unless we want to force a rerender in every case. Is there something we can key to?

jordanlesich: Come to think of it, I'd say we make this a generic slot that we can put any UI into.

jordanlesich (approved):
Just a few small notes.

Issue #417, opened 2022-07-12
TX Builder Consumes Arbitrary Application State


earth2travis:
## What

After a transaction, we need: 
- [x] All the context data passed as props into the TXBuilder Context 
- [x] formData
- [x] Keychain & endpoint data 
- [x] user/profile data
- [x] arbitrary data in TXlego 
- [x] arbitrary args from ReactComponent
- [x] Current dao, current chain, 
- [x] current daoType

all rolled up in one {}

A way to pass in optional typing for this data (ex. T extends Record<string, any> = Record<string, any> ) 

## Type System
- [x] Rebuid TXLego
- [x] Refactor all lego types to common utils

- [x] TXBuilder accepts and LifeCycle Fns as props and/or as arguments in FireTransaction()

## Refactor 
- [x] Refactor Summon to take in new arg structure
- [x] Refactor TX utils to handle new structure.

PR #598, opened 2022-08-03
Tx build/data aggregate

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/417

## Changes

Big list in issue

## Checks

- [x] App runs locally


apps/summon-app/src/utils/summonlegos.ts
```diff
@@ -0,0 +1,15 @@
+import { LOCAL_ABI } from '@daohaus/abi-utilities';
+import { ContractLego, CONTRACTS } from '@daohaus/common-utilities';
+
+const BaalSummonerContract: ContractLego = {
+  contractName: 'BaalSummoner',
+  type: 'local',
+  abi: LOCAL_ABI.BAAL_SUMMONER,
+  keychain: CONTRACTS.V3_FACTORY,
```
skuhlmann, line 8: nice just having the keychain here

libs/common-utilities/src/types/general.ts
```diff
@@ -2,3 +2,10 @@ export type Noun = {
   singular: string;
   plural: string;
 };
+
+// By definition this needs to use any.
+// But we should be dilligent about using this type and make sure that
+// in cases where we're using this to handle highly level functionality
+// we can extends with more specific types (ex. TXBuilder)
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type ArbitraryState = Record<string, any>;
```
skuhlmann, line 11: :+1:

libs/tx-builder-feature/src/utils/txBuilderUtils.ts
```diff
@@ -199,6 +212,7 @@ export async function handleFireTx({
     provider.getSigner().connectUnchecked()
   );
 
-  const ethersTx = await contract.functions[txName](...args);
-  return executeTx(tx, ethersTx, setTransactions, chainId);
+  const ethersTx = await contract.functions[method](...realArgs);
```
skuhlmann, line 215: Not needed right now, But I think this is likely were we want to send the gasLimit when we get to processProposal down the line.

Can use the overrides object:
https://docs.ethers.io/v5/api/contract/contract/#contract-functionsSend

skuhlmann (approved):

skuhlmann:
Looking good.

libs/tx-builder-feature/src/utils/txBuilderUtils.ts
jordanlesich: Thinking the same. Am I correct in assuming that this is only crucial for when we call processProposal?

skuhlmann: > Thinking the same. Am I correct in assuming that this is only crucial for when we call processProposal?

Yeah that is the only one we'll need it on.

Issue #455, opened 2022-07-17
Project wide lint and formatting settings not taking


brossetti1:
there was a conversation around formmatters in different devs environments overriding settings that come from nx. Hypothetically, nx should be setup so that we are all on the same formatting rules.

https://discord.com/channels/709210493549674598/946177876905037894/996559181412839477

unsure why this is currently but worth looking into.

brossetti1:
closing due to https://github.com/HausDAO/daohaus-monorepo/issues/6

PR #603, opened 2022-08-04
Default editor settings

brossetti1:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/455

## Changes

extract default settings into monorepo so format on save is unified between vs code and vim with project wide settings

if anyone has anything to add here via vim config, would highly appreciate a review

a +1 is the monthly dependabot task, runs monthly

## Packages Added

N/A

## Checks

- [ ] App runs locally

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: 'npm'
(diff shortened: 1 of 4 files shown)


jonathanprozzi (approved):
Nice!

Issue #626, opened 2022-08-08
Use Lens to Replace Ceramic


earth2travis:
# Overview

We are seeing some issues with ceramic library dependencies. In lieu of Ceramic's graph query implementation,  can we replace the ceramic profile fetch with lens profiles?

- Allows fetching on a single address
- Allows fetching on multiple addresses

Update the getProfile functions in dao-data and add one to fetch multiple addresses at once

https://docs.lens.xyz/docs/get-profiles

skuhlmann:
```gql
query Profiles {
  profiles(request: { ownedBy: ["0x83aB8e31df35AA3281d630529C6F4bf5AC7f7aBF"], limit: 10 }) {
    items {
      id
      name
      bio
      followNftAddress
      metadata
      isDefault
      
      picture {
        ... on NftImage {
          contractAddress
          tokenId
          uri
          verified
        }
        ... on MediaSet {
          original {
            url
            mimeType
          }
        }
        __typename
      }
      handle
      coverPicture {
        ... on NftImage {
          contractAddress
          tokenId
          uri
          verified
        }
        ... on MediaSet {
          original {
            url
            mimeType
          }
        }
        __typename
      }
      ownedBy


    }
    pageInfo {
      prev
      next
      totalCount
    }
  }
}

```

PR #630, opened 2022-08-08
Feature/lens

skuhlmann:
## GitHub Issue

[GitHub issue](https://github.com/HausDAO/daohaus-monorepo/issues/626).

## Changes

- swaps ceramic fetch for lens
- new lens graphql types and query
- propagates chanegs

## Packages Added

- none
## Checks

- [x] App runs locally


libs/dao-data/src/subgraph/queries-lens/profiles.generated.ts
```diff
@@ -0,0 +1,345 @@
...
+      id: any;
+      name?: string | undefined;
+      bio?: string | undefined;
+      followNftAddress?: any | undefined;
+      metadata?: any | undefined;
+      isDefault: boolean;
+      handle: any;
+      ownedBy: any;
+      picture?:
+        | {
```
brossetti1, line 21: never seen this syntax before, is the initial `|` needed here?

libs/dao-data/src/utils/transformers.ts
```diff
@@ -25,18 +25,21 @@ export const transformProposal = (
...
-  ens: string | null,
-  profile: BasicProfile
+  profile: LensProfile
 ): AccountProfile => {
   return {
     address,
-    ens,
-    ...profile,
+    name: profile?.name,
+    ens: profile?.onChainIdentity.ens?.name,
```
brossetti1, line 33: might want to add `?.ens` here in case onChainIdentity is empty

brossetti1 (approved):
left a few comments to check out, approving regardless

libs/dao-data/src/subgraph/queries-lens/profiles.generated.ts
skuhlmann: it's generated so i can't really control that.

Issue #573, opened 2022-07-31
Image Link Input


jordanlesich:
![Image](https://user-images.githubusercontent.com/53406838/182041028-cbc482a4-aeab-440d-a2f2-f24c24f023ce.png)

## Step 1: Review with Design
- What are we uploading to? We have no server or cache to store the image on. 
- I recommend making an image link input that displays the image once we add the link instead

## Step 2: Build
- If design ok's this pattern, then we compose WrappedInput and Avatar components
- Validate for valid link
- Display Image if link is valid

PR #654, opened 2022-08-12
Feature/image input

skuhlmann:
## GitHub Issue

[GitHub issue](https://github.com/HausDAO/daohaus-monorepo/issues/573).

## Changes

- adds an atom and wrapper components
- fileinput component, styles and story

fileinput:
Screenshot from 2022-08-15 15-41-23

wrapped input:
Screenshot from 2022-08-15 15-41-35

image input:
Screenshot from 2022-08-12 15-00-09
Screenshot from 2022-08-12 14-59-57

## Packages Added

n/a

## Checks

- [x] App runs locally


libs/common-utilities/src/utils/ipfs.ts
```diff
@@ -0,0 +1,24 @@
...
+}: {
+  creds: { pinata_api_key: string; pinata_api_secret: string };
+  formData: FormData;
+}) => {
+  try {
+    const response = await fetch(PINATA_PIN_FILE_API, {
+      method: 'POST',
+      headers: {
+        pinata_api_key: creds.pinata_api_key,
+        pinata_secret_api_key: creds.pinata_api_secret,
```
jordanlesich, line 16: How do devs get access to creds?

libs/ui/src/components/molecules/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <FieldWrapper {...props} id={id}>
+      <FileInputContents>
+        <Button
+          onClick={() => {
+            handleBrowse();
+          }}
+        >
+          {buttonText}
+        </Button>
+        <input
```
jordanlesich, line 57: We can wrap WrappedInput. No need to remake a new input

jordanlesich (commented):
Perhaps we could take some time today to discuss a couple of the notes I made. 

As with RPCs and Explorer keys, we'll need to establish a strategy where devs can add their own keys to get this working. 

Also, we need to be using currently existing Wrapped components so that they stay updated once we make changes to the base component.

libs/common-utilities/src/utils/ipfs.ts
skuhlmann: devs need to pass their own creds

libs/ui/src/components/molecules/FileInput/FileInput.tsx
skuhlmann: It didn't seem like WrappedInput supported the file type input

jordanlesich: I don't think it will. We can make sure that the input doesn't recieve that data though. I can pair up and show what I have in mind if you have time.

libs/common-utilities/src/utils/ipfs.ts
jordanlesich: For sure. What do we have in mind for comm devs to be able to have the own creds, existing env var process (not necessarily opposed to this.

jordanlesich (approved):
The code looks good to me. The only thing I struggle with is the look (since it's strayed a lot from the original). 

Would it make sense to quickly meet with design and punch it up a tad?

Issue #629, opened 2022-08-08
Create GatedCheckbox Component


skuhlmann:
# Details

We need a way to wrap an input as 'Gated' so if a user checks the checkbox they see the form input/area.

Includes:
- checkbox
- takes text for the explaination
- if checked it shows the input - if not it hides it
- takes the input as a prop

TODO: How does this need to play with form builder?
@jordanlesich 

Example usage in the expiration date component:

![Image](https://user-images.githubusercontent.com/6923345/183709284-1adf917f-79fb-4794-95ac-601adcfe37eb.png)

jordanlesich:
Prop is perfect. That way we can define the gate as a lego. 

We should test it to make sure that it works with other registered Inputs.

santteegt:
two questions:

- gated Input should support most of the wrapped input field components (molecules) ? Or should also support atom components?
- Explanation should be a markdown text?

Issue #498, opened 2022-07-21
Add Expiration Date Form Component


jordanlesich:
![Image](https://user-images.githubusercontent.com/42706432/183456912-866684e4-406e-4d5d-814b-e252645b6283.png)

https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6120%3A132023

Component that lets users pick the expiration date of their proposal.

See PL components > Input Molecules for this interaction. 

Expected behavior: 
- 'Add expiration date' single checkbox appears above the submit button on every proposal 
- When clicks, the same input as for selecting proposal timing (grace, voting) periods is used 
- This defaults to 0 days
- The time input into this field is added to the DAOs voting and grace period
- When a user updates the field (asap on user input - can be when 1 or 2 seconds user stops changes in the input) the date/time in the text below updates 
- An input of some kind if required, they can't leave it blank - error message 'Please input a length of time or uncheck the add expiration date checkbox' 
- User can uncheck the box at any time and the field + text goes back to before, the field is cleared

earth2travis:
@adriennetilley I moved this to Invisible Suburbs

skuhlmann:
@santteegt

PR #660, opened 2022-08-15
Feature/Gated Input

santteegt:
## GitHub Issue

#629 

## Changes

Introduces a new component that includes a toggle checkbox  to show/hide an input molecule

## Packages Added

None

## Checks

- [ ] App runs locally

### DIFF
--- a/libs/ui/src/components/molecules/GatedInput/GatedInput.stories.tsx
+++ b/libs/ui/src/components/molecules/GatedInput/GatedInput.stories.tsx
@@ -49,7 +49,7 @@ FullGatedInput.args = {
   error: undefined,
   success: undefined,
   input: WrappedInput,
-  updateHelperMsg: (formValues) => `This is the new value: ${formValues['example']}`,
+  updateHelperMsg: (formValues) => `This is the new value: ${formValues}`,
   children: defautExplanationText,
 };
 
@@ -71,7 +71,8 @@ FullGatedInputSelect.args = {
   selectPlaceholder: '-period-',
   input: WrappedInputSelect,
   updateHelperMsg: (formValues) =>
-    `This is the new value: ${formValues['selectID'] ? formValues['inputId'] * formValues['selectID'] : 0} seconds`,
+    `This is the new value: ${formValues.length ? formValues[0] * formValues[1] : 0} seconds`,
+  customWatchers: ['inputId', 'selectID'],
   children: (
     <ParXs style={{marginTop: '5px'}}>
       The expiration Date includes Voting and Grace periods. Adjust the days or hours
@@ -98,7 +99,7 @@ FullGatedSelect.args = {
   warning: undefined,
   error: undefined,
   input: WrappedSelect,
-  updateHelperMsg: (formValues) => `This is the new value: ${formValues['example'] || ''}`,
+  updateHelperMsg: (formValues) => `This is the new value: ${formValues || ''}`,
   children: defautExplanationText,
 };
 
@@ -138,6 +139,6 @@ FullGatedTextArea.args = {
   error: undefined,
   success: undefined,
   input: WrappedTextArea,
-  updateHelperMsg: (formValues) => `This is the new value: ${formValues['example'] || ''}`,
+  updateHelperMsg: (formValues) => `This is the new value: ${formValues || ''}`,
   children: defautExplanationText,
 };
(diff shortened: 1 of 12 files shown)


brossetti1 (approved):
this looks good to me - served up the storybook - lots of good examples

libs/ui/src/components/molecules/GatedInput/GatedInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+  updateHelperMsg?: (value: FieldValues) => string; 
+  children?: ReactNode;
+};
+
+export const GatedInput = ({ children, gatedId, gateLabel, input: Input, ...props }: Buildable<GatedInputProps>) => {
+  const { watch } = useFormContext();
+  const { updateHelperMsg } = props;
+  const [gatedOn, toggleGate]= useState(false);
+  const [helperText, setHelperText] = useState(props.helperText || '');
+  const inputValues = watch();
```
jordanlesich, line 22: Can we can scope this to the actual input that input is watching?

const inputValue = watch(id)

libs/ui/src/components/molecules/GatedInput/GatedInput.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { FieldWrapper } from '../FieldWrapper';
+import { Checkbox } from '../../atoms/Checkbox';
+import { Buildable, IGatedInput } from '../../../types/formAndField';
+import { GatedInputBox } from './GatedInput.styles';
+
+type GatedInputProps = IGatedInput & {
+  gatedId: string;
+  gateLabel: string;
+  input: FunctionComponent<Buildable<IGatedInput>>;
+  updateHelperMsg?: (value: FieldValues) => string;
```
jordanlesich, line 13: What is the use case for passing in a callback for generating helper text?

libs/ui/src/components/molecules/GatedInput/GatedInput.stories.tsx
```diff
@@ -0,0 +1,143 @@
...
+  options: [
+    { value: 'selectValue1', name: 'Select Label 1' },
+    { value: 'selectValue2', name: 'Select Label 2' },
+  ],
+  long: false,
+  full: false,
+  warning: undefined,
+  error: undefined,
+  input: WrappedSelect,
+  updateHelperMsg: (formValues) => `This is the new value: ${formValues['example'] || ''}`,
```
jordanlesich, line 101: Let's shy away from using callbacks to describe how we want our field elements to behave. It makes serialising the lego data difficult.

jordanlesich (changes requested):
Hey Santiago, 

Let's chat about this helperText API some time. There's a couple issues I'm seeing here that I think we could find some easy solutions for,

libs/ui/src/components/molecules/GatedInput/GatedInput.tsx
santteegt: InputSelect is an example where you might need access to values in both input and select components

libs/ui/src/components/molecules/GatedInput/GatedInput.tsx
santteegt: The Expiration Date component https://github.com/HausDAO/daohaus-monorepo/issues/498 requires updating the helper text based on the input values. I could set this text as children but I'm wondering where exactly this explanation text updates should be handled later in the form

image

libs/ui/src/components/molecules/GatedInput/GatedInput.stories.tsx
santteegt: same concerns from above about how to update the explanation text

libs/ui/src/components/molecules/GatedInput/GatedInput.tsx
jordanlesich: Better to do const [inputVal, selectVal] = watch([inputId, selectId])

libs/ui/src/components/molecules/GatedInput/GatedInput.stories.tsx
jordanlesich: I'd really like to not have a component that takes in a callback as a means to produce helper text. There's a good pattern we can use to get around this. DM me on Discord when you're available to pair on this.

santteegt:
superseded by #690

Issue #574, opened 2022-07-31
Shaman Permissions Input


jordanlesich:
## What 

![Image](https://user-images.githubusercontent.com/53406838/182041273-735b4c2b-4d03-421a-9ac6-b130502924ec.png)

- Select that Displays Shaman Data when a permission level is selected
- The value is just the number listed below

option content:

- 0 - Remove All Permissions
- 1 - Admin Only
- 2 - Manager Only
- 3 - Admin and Manager
- 4 - Governance Only
- 5 - Admin and Governance
- 6 - Manager and Governance
- 7 Admin and Manager and Governance

Issue #568, opened 2022-07-31
Add Modifier Button to Field Wrapper


jordanlesich:
### Update Field wrapper and button 
![Image](https://user-images.githubusercontent.com/53406838/182004423-0771e883-3cc1-41cf-a2a9-1362ed0d224b.png)
- Add button to fieldWrapper and utilButton Prop
- Ensure button and helper text are part of the same container
- Helper text wraps as needed. 

**If** Design thinks we are putting other components where the button is, then we make a modifer slot and pass a button down to it
**Else** We pass an object {label: string, onClick: () => } to a prop called 'ModButton'

skuhlmann:
Do the 1st option - let dev pass anything down.

earth2travis:
Let's roll with making a modifier slot and pass a button down to it

earth2travis:
Constrained width is a bit tricky

skuhlmann:
@santteegt

PR #665, opened 2022-08-15
Add shaman permission input

alexkeating:
## GitHub Issue

#574 

## Changes

Add a shaman permission input to core app


libs/common-utilities/src/constants/contracts.ts
```diff
@@ -36,3 +36,38 @@ export const CONTRACTS: ContractAddressList = {
     '0x5': '0x000000000000cd17345801aa8147b8d3950260ff',
   },
 };
+
```
skuhlmann, line 39: see unused import on this one.

apps/core-app/src/components/ShamanInput.tsx
```diff
@@ -0,0 +1,10 @@
+import { Select } from '@daohaus/ui';
```
skuhlmann, line 1: Let's move this into the UI library as a moecule

apps/core-app/src/components/ShamanInput.tsx
```diff
@@ -0,0 +1,10 @@
...
+import { SHAMAN_PERMISSIONS } from '@daohaus/common-utilities';
+
+export const ShamanInput = () => {
+  const options = Object.values(SHAMAN_PERMISSIONS).map((permission) => ({
+    name: permission.displayName,
+    value: permission.id,
+  }));
+  return <Select id="shaman-input" options={options} />;
```
skuhlmann, line 9: Use WrappedSelect here

libs/ui/src/types/formAndField.ts
```diff
@@ -36,14 +36,17 @@ export type PrimitiveSizable = {
 };
 
 export type PrimitiveAddons = {
-  rightAddon?: React.ReactNode
-}
+  rightAddon?: React.ReactNode;
+};
```
skuhlmann, line 40: Thinking we might not want to extend right add on to all fields?

skuhlmann (changes requested):

libs/ui/src/types/formAndField.ts
alexkeating: This is only a formatting change change, there should be changed from what already exists

libs/common-utilities/src/constants/contracts.ts
alexkeating: Are you referring to the first line of the file? I can remove that

apps/core-app/src/components/ShamanInput.tsx
alexkeating: Will update

apps/core-app/src/components/ShamanInput.tsx
alexkeating: Will do

libs/ui/src/types/formAndField.ts
skuhlmann: > This is only a formatting change change, there should be changed from what already exists

ah whoops. got it.

jordanlesich: Let's figure out where that got put in and address it.

skuhlmann: > Let's figure out where that got put in and address it.

@jordanlesich 

It was added with this one that says to add it to fieldwrapper:
https://github.com/HausDAO/daohaus-monorepo/issues/568

Where would you rather see it? I'll open a new ticket to move it.

jordanlesich: Oh, it's just for the wrapper. I thought it was for the input. My bad. This is exactly what we want.

libs/common-utilities/src/constants/contracts.ts
alexkeating: I removed the unused import above

alexkeating:
Hopefully should be good now! Thanks for the review

skuhlmann (approved):

Issue #541, opened 2022-07-29
Summon: adding decimals to fields can cause form errors


adriennetilley:
### Feature

summon

### Expected outcome

When adding decimals to fields, it should either:
- allow it, ex. New Offering should allow for .01 ETH 
- not allow it, ex. .1 seconds is not an option

### Actual outcome

When adding seconds to some fields as error is thrown that won't go away until you refresh the page. 

### Browser

Chrome

### Screenshots and context

![sPykT37](https://user-images.githubusercontent.com/42706432/181757120-877cac2f-edd9-4815-a0fc-ca5c4d03ab95.png)

earth2travis:
![Image](https://user-images.githubusercontent.com/522182/179835395-c6126ca1-2d7f-43a5-9489-3b6f77cae5ce.png)

![Image](https://user-images.githubusercontent.com/522182/179835427-34da7e05-cdff-4a24-b674-fa5d24c00570.png)

PR #677, opened 2022-08-16
Fix/summon decimals

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/541

## Changes

- Fixed decimal bug
- Added modifier to ensure that value is converted to base units
- Added validators for all the other number components in that segment
- created percentage validator (limits to range of 1-100 or custom range) in the Validators object
- added percentage validator to Quorum field 

## Packages Added
NA

## Checks

- [x] App runs locally


apps/summon-app/src/utils/summonTx.ts
```diff
@@ -52,6 +52,8 @@ const governanceConfigTX = (formValues: FormValues) => {
       'governanceConfigTX recieved arguments in the wrong shape or type'
     );
   }
+
+  console.log('newOffering', newOffering);
```
skuhlmann, line 56: tidy up

skuhlmann (approved):
looks good

PR #683, opened 2022-08-16
Form/signal proposal

jordanlesich:
## GitHub Issue

https://github.com/orgs/HausDAO/projects/18

## Changes

- Sets up builders for core app (file creation, organization, custom types) 

## Packages Added
NA

## Checks

- [x] App runs locally

### DIFF
--- a/apps/core-app/src/legos/form.ts
+++ b/apps/core-app/src/legos/form.ts
@@ -1,5 +1,6 @@
 import { CustomFormLego } from './config';
 import { FIELD } from './fields';
+import { TX } from './tx';
 
 export const FORM: Record<string, CustomFormLego> = {
   SHARE_SWAP: {
@@ -16,6 +17,7 @@ export const FORM: Record<string, CustomFormLego> = {
     subtitle: 'This is a test form',
     description: 'This is a test form',
     requiredFields: { title: true, description: true },
+    tx: TX.POST_SIGNAL,
     fields: [FIELD.TITLE, FIELD.DESCRIPTION, FIELD.LINK],
   },
 };
--- a/apps/core-app/src/legos/form.ts
+++ b/apps/core-app/src/legos/form.ts
@@ -1,0 +1,21 @@
+import { CustomFormLego } from './config';
+import { FIELD } from './fields';
+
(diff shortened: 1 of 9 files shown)


apps/core-app/src/legos/tx.ts
```diff
@@ -0,0 +1,28 @@
...
+          {
+            type: 'JSONDetails',
+            jsonSchema: {
+              title: `.formValues.title`,
+              description: `.formValues.description`,
+              link: `.formValues.link`,
+              proposalType: { type: 'static', value: 'Signal Proposal' },
+            },
+          },
+          { type: 'static', value: POSTER_TAGS.daoProfileUpdate },
```
jordanlesich, line 23: Just realized that this isn't a good idea. We'll need a tag for this.

skuhlmann: ok yeah and we need an agreed upon schema. let think about that.

just title, description and link? we'll also need daoid in there.

skuhlmann: and we'll need to do the subgraph stuff - but good to get some work around that.

let's talk later on it

apps/core-app/src/legos/form.ts
```diff
@@ -0,0 +1,23 @@
...
+    fields: [FIELD.TITLE, FIELD.DESCRIPTION, FIELD.LINK],
+  },
+  SIGNAL: {
+    id: 'test',
+    title: 'Test Form',
+    subtitle: 'This is a test form',
+    description: 'This is a test form',
+    requiredFields: { title: true, description: true },
+    tx: TX.POST_SIGNAL,
+    fields: [FIELD.TITLE, FIELD.DESCRIPTION, FIELD.LINK],
```
skuhlmann, line 21: does it need an expiration too? or should singal default to no expiration?

jordanlesich: So for expiration currently, seeing as we're awaiting the input, I have it so that it will search the value automatically from any value '.formData.expiry'. And if there isn't a value, it will default to two weeks from proposal submit. I can also default to 0 as we discussed earlier on Monday.

skuhlmann (approved):

PR #692, opened 2022-08-20
Fix modifier button positioning on FieldWrapper

santteegt:
## GitHub Issue

Related to #568 

## Changes

Fix modifier button positioning within FieldWrapper to be fixed at the right

## Packages Added

None

## Checks

- [X] App runs locally

### DIFF
--- a/libs/ui/src/components/molecules/FieldWrapper/FieldWrapper.styles.ts
+++ b/libs/ui/src/components/molecules/FieldWrapper/FieldWrapper.styles.ts
@@ -32,8 +32,6 @@ export const LabelContainer = styled.label`
 
 export const BottomContainer = styled.div`
   display: flex;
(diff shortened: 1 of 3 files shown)


libs/ui/src/components/molecules/FieldWrapper/FieldWrapper.styles.ts
```diff
@@ -32,8 +32,7 @@ export const LabelContainer = styled.label`
 
 export const BottomContainer = styled.div`
   display: flex;
-  align-items: center;
-  justify-content: space-between;
+  flex-direction: row;
```
0xRowdy, line 35: Don't need this row is default.
```suggestion
```

0xRowdy (approved):

Issue #586, opened 2022-08-02
Create Expand Card Component


adriennetilley:
Ticket for the card expander: 

![Image](https://user-images.githubusercontent.com/42706432/182430621-7c0d9cb4-839f-4a34-9c1d-a12699c3455d.png)

Behavior:
- Show/hide button always in bottom-right side of top card
- Clicking expands out more detailed information below - this content can be anything aside from another top-level card
- Ideally this slide animates downward, then content fades in quickly (~300ms or default) the content

adriennetilley:
Ready for spec

earth2travis:
@alexkeating if you have some spare cycles this week let me know and we can bring this over

PR #693, opened 2022-08-21
586 expandable card

0xRowdy:
## GitHub Issue

[GitHub issue.](https://github.com/HausDAO/daohaus-monorepo/issues/586)

## Changes

Created a Card with a trigger that allows the card to display the hidden content.

image

## Packages Added

@radix-ui/react-collapsible

## Checks

- [x] App runs locally


libs/ui/src/components/atoms/Tooltip/Tooltip.styles.ts
```diff
@@ -9,7 +9,7 @@ import { border } from '../../../theme/global/border';
 export const TooltipProvider = TooltipPrimitive.Provider;
 export const TooltipRoot = TooltipPrimitive.Root;
 export const TooltipTrigger = styled(TooltipPrimitive.Trigger)`
-  background-color: ${({ theme }: { theme: Theme }) => theme.bgColor};
+  background-color: transparent;
```
brossetti1, line 12: just checking - was this suppose to be changed?

0xRowdy: Yeah, the background was colored with the primary theme. It made it so there was a small box that was a different color than the card. It makes more sense to have the background be the color of the parent.

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+
+export const StyledCollapsibleContent = styled(Collapsible.Content)`
+  /* TODO Animations? */
+  /* &[data-state='open'] {
+    animation: ${open} 300ms ease-out;
+  }
+  &[data-state='closed'] {
+    animation: ${close} 300ms ease-out;
+  } */
+`;
```
brossetti1, line 60: seems like a good idea, the animation here is simple and goes along way to making interactions more pleasnt

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+  onChange,
+  width = 'fit-content',
+  children,
+  collapsibleContent,
+}: CardProps) => {
+  return (
+    <StyledCollapsibleCard open={open} onChange={onChange} width={width}>
+      <InnerCard width={'90%'}>
+        {children}
+        <StyledCollapsibleCardTrigger asChild>
```
brossetti1, line 31: is `asChild` something in our library or is that coming from radix?

libs/ui/src/styled.d.ts
```diff
@@ -68,6 +68,16 @@ declare module 'styled-components' {
...
+    collapsibleCard: {
+      bg: string;
+      bgInnerCard: string;
+      border: string;
+      borderInnerCard: string;
+      hoverBg: string;
+      hoverBorder: string;
+      focusBg: string;
+      focusBorder: string;
+    };
```
brossetti1, line 80: so not sure where this config comes from, only ones I am seeing being used are `borderInnerCard` and `bgInnerCard`

looking in the file i see some of these repeated so not sure what the purpose is - are the unused attributes for custom themeing for the daos in the ui?

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.styles.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const open = keyframes({
+  from: { height: 0 },
+  to: { height: 100 },
+});
+
+const close = keyframes({
+  from: { height: 100 },
+  to: { height: 0 },
+});
```
brossetti1, line 50: should these go into https://github.com/HausDAO/daohaus-monorepo/blob/develop/libs/ui/src/animations/general.ts#L15?

brossetti1 (approved):
left a few questions, mainly getting used to the UI library now. Not the best at css so cant help out there too much -- looks good otherwise

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.styles.ts
0xRowdy: Good call.

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.tsx
```diff
@@ -0,0 +1,54 @@
...
+          <StyledCollapsibleCardButton>
+            Expand Me
```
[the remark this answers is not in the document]
0xRowdy: No this should be a prop. Ty

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.tsx
0xRowdy: asChild is implemented by radix. It allows for us to override the element radix uses. In this example redux uses buttons for triggers. We could use a div or SVG or whatever. We use a button but I don't want a button inside a button or to restyle the redux trigger button.

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.styles.ts
0xRowdy: Fixed

libs/ui/src/styled.d.ts
0xRowdy: Cleaned up

libs/ui/src/components/molecules/CollapsibleCard/CollapsibleCard.styles.ts
0xRowdy: Finished

Issue #572, opened 2022-07-31
Create Base Warning Atom and UnlockToken Molecule Components


jordanlesich:
### Base UI

- Build the Base Field Alert component (can be named something else) 
- Displays a message (left aligned) and arbitrary element (right aligned)
- Consumes our Haus theme, 
- can display success, warning, error states 

![Image](https://user-images.githubusercontent.com/53406838/182040876-0bf6ca3e-99a3-411d-9a25-435634ee64a1.png)

### Unlock Token Molecule
- Fires Unlock token 
- Composes Base UI

earth2travis:
Jord said this one was built in TributePicker

PR #704, opened 2022-08-22
Warning molecule

brossetti1:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/572

- I decided to just add the Field Alert into molecules - didnt feel right in atoms
- wondering if this can just be `Alert` as well as children (button) are optional?

## Changes

- [X] adding `FieldAlert` Atom
- [X] adding stories for `FieldAlert` 
- [X] adding `Unlock Token Molecule`

## Packages Added

N/A

## Checks

- [X] App runs locally

## Details

Screen Shot 2022-08-23 at 12 44 53 PM
Screen Shot 2022-08-23 at 12 44 59 PM
Screen Shot 2022-08-23 at 12 45 03 PM


apps/core-app/src/components/customFields/tributeInput.tsx
```diff
@@ -296,11 +272,13 @@ const TemporaryWarning = ({
...
+    <FieldAlert
+      className="warning"
+      message={`You must approve ${tokenName || 'Token'} to submit`}
+    >
+      <Button sm type="button" onClick={handleApprove}>
         {txState === TxStates.Loading ? 'Loading...' : 'Approve'}
-      </WarningButton>
-    </TempWarningBox>
+      </Button>
+    </FieldAlert>
```
brossetti1, line 282: one thing im curious about here is if this is how we would want a `FeildAlert` to act - basically if you pass it a `ReactNode`, that ReactNode will be presented on the Right side of the component.

libs/ui/src/components/molecules/FieldAlert/FieldAlert.styles.tsx
```diff
@@ -0,0 +1,58 @@
...
+      :hover {
+        background-color: ${({ theme }: { theme: Theme }) => theme.fieldAlert.warningButtonBg};
+        border: 1px solid ${({ theme }: { theme: Theme }) => theme.fieldAlert.warningButtonBorder};
+      }
+      :active {
+        background-color: ${({ theme }: { theme: Theme }) => theme.fieldAlert.warningButtonBg};
+        border: 1px solid ${({ theme }: { theme: Theme }) => theme.fieldAlert.warningButtonBorder};
+      }
+    }
+  }
```
brossetti1, line 45: having to override warning for p and button on this component, as the current ui library does not have these warning colors set.

apps/core-app/src/components/customFields/tributeInput.tsx
jordanlesich: Interesting. In these situations, I generally use prop names that describe where things are going to go  (ex. renderLeft or renderRight) but this works too.

libs/ui/src/components/molecules/FieldAlert/FieldAlert.styles.tsx
jordanlesich: Button isn't ideal in this situation, but we're going to overhaul the button system soon, so let's leave it in for now. We'll have the utility button types later on.

jordanlesich:
Just noticed the border on the primary field alert is blue. Text shouldn't be white either.

Issue #624, opened 2022-08-08
Proposals: Build Manage/Update Forms for Shamans


earth2travis:
This ticket covers the 'Manage Shaman' and 'Update Shaman' proposal forms.

### Add Shaman
Designed layout here: https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6209%3A134780

### Manage Shaman
Designed layout here: https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6216%3A137653

## Proposal Data
Type: Shaman Proposal
Proposal info that will appear on the proposal details: 'This proposal is adding a Shaman with admin permissions to the DAO. Be sure to look at the contract details for more information.'

### Add Shaman Form Fields
- Shaman Proposa (subheader)
- Add a Shaman (h2)
- Learn more about Shamans in our documentation. (instructional text, link goes to https://baal-docs.vercel.app/tools/shaman)
- Basics (should be the same on all proposals:
    - Proposal Name (required, text field)
    - Description (text area)
    - Link (placeholder 'http://', text field)
 - Shaman Address (required, address input validation) 
 - Permission Level (required, dropdown, default to 0 admin level) 
     - 0 - No Permissions
     -  1 - Admin Only
     -  2 - Manager Only
     -  4 - Governance Only
     -  3 - Admin + Manager
      - 5 - Admin + Governance
      - 6 - Manager + Governance
      - 7 - Admin + Manager + Governance 
- Add Expiration - https://github.com/HausDAO/daohaus-monorepo/issues/498
- 'Submit Proposal' > transaction > takes user to proposal list

### Manage Shaman Form Fields
- Shaman Proposa (subheader)
- Manage Shaman (h2)
- Learn more about Shamans in our documentation. (instructional text, link goes to https://baal-docs.vercel.app/tools/shaman)
- Basics (should be the same on all proposals:
    - Proposal Name (required, text field)
    - Description (text area)
    - Link (placeholder 'http://', text field)
- Shaman (disabled, has current shaman address we are adjusting in place)
- New Permission Level (dropdown, same as Permission Level field above, should default to current shaman's permission)
- Shaman's old permission level is: PermissionValue (text)
- Shaman's new permission level is: UserInputForNewLevel (text, updates with dropdown update to previous question
- Add Expiration - https://github.com/HausDAO/daohaus-monorepo/issues/498
[...]

PR #723, opened 2022-08-25
Shaman forms

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/624

## Changes

- Add shaman form 
- manage Shaman form 
- Shaman deluxe 
- Fix field validation empty, non-required fields
- Fix search throwing for falsy "" search results
- Fix disabled field override from formBuilderFactory 

## Packages Added

## Checks

- [x] App runs locally


apps/core-app/src/components/customFields/ShamanDeluxe.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const newRules: RegisterOptions = {
+    validate: (val) =>
+      ignoreEmptyVal(val, (val) => {
+        if (oldShamanLevel == null) return true;
+        return Number(oldShamanLevel) > Number(val)
+          ? true
+          : 'Shaman permission level can only go down';
+      }),
+    ...props.rules,
```
brossetti1, line 54: should this technically be above `validate` so it does not accidentally override validate in the rules?

libs/ui/src/components/molecules/index.ts
```diff
@@ -1,12 +1,14 @@
 export * from './AddressDisplay';
 export * from './CSInput';
+export * from './CheckGate';
```
brossetti1, line 3: looks like a duplicate here

apps/core-app/src/components/customFields/ShamanDeluxe.tsx
jordanlesich: I've been experimenting with spreading the rules after the declaration so that wrapper components can easily overwrite validation rules if need be.

libs/ui/src/components/molecules/index.ts
jordanlesich: Fixed.

skuhlmann (approved):

PR #742, opened 2022-08-26
fixing json error

brossetti1:
## GitHub Issue

no linked issues - fixing settings page load issue with `dao?.links`

## Changes

links are coming in as an object instead of a string, handling them accordingly

## Packages Added

N/A

## Checks

- [ ] App runs locally


libs/dao-data/src/types/query.types.ts
```diff
@@ -122,7 +122,7 @@ export type DaoProfile = {
   longDescription?: string;
   avatarImg?: string;
   tags?: string[];
-  links?: string;
+  links?: { [field: string]: string };
```
skuhlmann, line 125: I still want to test this on some other daos - i feel like it might come through as a string on stuff added from our new form.

skuhlmann (approved):

Issue #548, opened 2022-07-29
Update DAO Card Tag Colors


jonathanprozzi:
### Feature

Exploring the Hub as a connected user in at least 1 DAO

### Expected outcome

This is working as intended, but it's a minor enhancement -- we can utilize more dynamic colors for our tags (or perhaps different tag colors for different pieces of information)

### Actual outcome

Currently, all of the tags used in the DAO Cards are the same shade of blue and don't have a lot of contrast. We may want to explore how we want to generate tag colors.

### Browser

Brave

### Screenshots and context

<img width="399" alt="image" src="https://user-images.githubusercontent.com/9438776/181770165-3d43200b-7757-4517-8529-95e8354785d1.png">

earth2travis:
@adriennetilley do we want to change the shade of these tags?

earth2travis:
Is the problem that all of the tags are the same color?

The only one that seems to need to draw attention is delegate. Should we create a badge for delegate rather than using a tag?

We also need to refactor the tag component as it is not using our theming system.

adriennetilley:
Sorry for the wait on this!

Let's do the following color updates:
- Network in radix red - bg: dark/3, text: dark/11
- Delegate tag changed to radix amber - bg: dark/amber, text: dark/11
- Contract name in radix indigo (kept the same as now) 
<img width="332" alt="Screen Shot 2022-08-19 at 1 12 13 PM" src="https://user-images.githubusercontent.com/42706432/185672375-292a811f-21e5-4944-93d3-bd2654ecf201.png">

PR #745, opened 2022-08-26
adding tag colors

brossetti1:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/548

## Changes

- [X] adding tag colors

Screen Shot 2022-08-29 at 12 17 05 PM
Screen Shot 2022-08-29 at 12 17 09 PM
Screen Shot 2022-08-29 at 12 17 25 PM

## Checks

- [ ] App runs locally


apps/hub-app/src/components/DaoCard.tsx
```diff
@@ -129,7 +129,7 @@ export const DaoCard = ({
         )}
       </div>
       <div className="tag-box">
-        <Tag>{networkId}</Tag>
+        <Tag className="network">{networkId}</Tag>
```
0xRowdy, line 132: There is a Tag component in the component lib.
It might need a little refactor to align with the different types but shouldn't be too bad.

Currently the tag component is set up to do take the prop tagColor: 'blue' | 'green' | 'pink' | 'violet';
It would be better to define the specific variants of a tag and pass those as props (ex. 'network' | 'delegate' | etc...). Then you can apply those either dynamically or create class names out of the prop.

apps/hub-app/src/components/Tag.tsx
```diff
@@ -1,17 +1,29 @@
 // COMPONMENT LIBRARY
 
 import styled from 'styled-components';
-import { blueDark } from '@radix-ui/colors';
+import { blueDark, redDark, yellowDark } from '@radix-ui/colors';
```
0xRowdy, line 4: Yes, we're trying to bring in all of the colors in the component library. It is in theme/global/colors I believe when I made the tag component it was a rush to get it in before a demo so I didn't actually create the component-specific alias in our theme yet.

It is some tech debt that we should address.

0xRowdy (commented):
There is a Tag component in the UI Library. 
Why not use that?

Happy to hop on a call to discuss

jordanlesich (approved):

Issue #717, opened 2022-08-24
Base proposal card


skuhlmann:
# Details

component in the core ui

![Image](https://user-images.githubusercontent.com/6923345/186530002-71701e60-555e-4bea-b545-d2dd0ae0284b.png)

Lefthand details:
- type
- createdon
- name
- description
- membercard - submitter
- link to details

Issue #714, opened 2022-08-24
Proposal list filter, search and load more


skuhlmann:
- Scour each proposal ticket and see if we can get all this data to the view
- Issue will be the logged in user vote data - will need an extra fetch for each proposal in the list
- also think about this for the proposal detail view
   - dao data needs a findProposal fetch with the logged in members data
   - check that vote totals are available so we don't have to loop through all nested votes

PR #751, opened 2022-08-29
Feature/714 proposal list

skuhlmann:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/714
fixes https://github.com/HausDAO/daohaus-monorepo/issues/717

## Changes

- prop card and list components
- search and filter and pagination hookup
- dao data updates to filter helpers

## Checks

- [x] App runs locally


apps/core-app/src/pages/Proposals.tsx
```diff
@@ -1,43 +1,130 @@
...
-  const { proposals } = useProposals();
+  const {
+    proposals,
+    setProposalsPaging,
+    proposalsNextPaging,
+    setProposalsFilter,
+    setProposals,
+  } = useProposals();
+  const { dao } = useDao();
+  const [searchTerm, setSearchTerm] = useState<string | ''>('');
```
0xRowdy, line 45: I think that empty quotes would be considered a string. So below should work:
```
  const [searchTerm, setSearchTerm] = useState<string>('');
```
or
```
const [searchTerm, setSearchTerm] = useState<string | null>(null);
```

apps/core-app/src/pages/Proposals.tsx
```diff
@@ -1,43 +1,130 @@
...
+
+    const filterQuery =
+      filter !== ''
+        ? statusFilter(
+            PROPOSAL_STATUS[filter],
+            Number(dao?.votingPeriod) + Number(dao?.gracePeriod)
+          )
+        : undefined;
+
+    if (debouncedSearchTerm && debouncedSearchTerm !== '') {
```
0xRowdy, line 66: I believe those checks are the same. An empty string is always false as a boolean

Maybe check against the length of the string? If you decide to have the initial state as null you would want to check against that also.
```suggestion
    if (debouncedSearchTerm && debouncedSearchTerm.length > 0) {
```

apps/core-app/src/utils/constants.ts
```diff
@@ -0,0 +1,10 @@
+export const PROPOSAL_FILTERS: { [key: string]: string } = {
```
0xRowdy, line 1: Use an Enum here?

apps/core-app/src/pages/Proposals.tsx
```diff
@@ -1,43 +1,130 @@
...
+        ...filterQuery,
+        title_contains_nocase: debouncedSearchTerm,
+      });
+      setProposalsPaging(defaultDaoData.proposalsPaging);
+    } else {
+      setProposalsFilter(filterQuery);
+      setProposalsPaging(defaultDaoData.proposalsPaging);
+      setProposals(undefined);
+    }
+    // TODO: I don't want to fire on these others!!
```
0xRowdy, line 78: I believe you called this out in standup today. Do you only want to fire on debouncedSearchTerm?

0xRowdy (approved):

apps/core-app/src/pages/Proposals.tsx
```diff
@@ -1,43 +1,130 @@
...
+
+  const toggleFilter = (event: MouseEvent<HTMLButtonElement>) => {
+    const searchQuery =
+      searchTerm !== '' ? { title_contains_nocase: searchTerm } : undefined;
+    setFilter((prevState) => {
+      if (prevState === event.currentTarget.value) {
+        setProposalsFilter(searchQuery);
+        setProposalsPaging(defaultDaoData.proposalsPaging);
+        setProposals(undefined);
+        return '';
```
jordanlesich, line 96: Curious what this function is returning values to,

jordanlesich (approved):

jordanlesich:
Approved. A couple minor tweaks to be made. Maybe some useCallbacks depending on the return of that function that I highlighted.

apps/core-app/src/components/FilterDropdown.tsx
```diff
@@ -0,0 +1,68 @@
...
+              className={filter === filterKey ? 'selected' : ''}
+            >
+              {PROPOSAL_FILTERS[filterKey]}
+            </DropdownButton>
+          </DropdownMenuItem>
+        );
+      })}
+    </Dropdown>
+  );
+};
```
brossetti1, line 66: this is really clean, nice!

apps/core-app/src/components/MetadataSettings.tsx
```diff
@@ -55,14 +54,17 @@ type MetadataSettingsProps = {
...
       <MetaCardHeader>
         <H3>Metadata</H3>
-        <Link href={`/molochv3/${daochain}/${daoid}/settings/update`}>
-          <Button>Update Settings</Button>
-        </Link>
+        {userMembership && Number(userMembership.shares) && (
+          <Link href={`/molochv3/${daochain}/${daoid}/settings/update`}>
+            <Button>Update Settings</Button>
+          </Link>
+        )}
```
brossetti1, line 67: ah thanks for cleaning this up

apps/core-app/src/pages/Proposals.tsx
brossetti1: i think its reseting to default to no filtering -- ive seen it in a few places, maybe it could be pulled out into its own function?

apps/core-app/src/utils/debounceHook.ts
```diff
@@ -0,0 +1,17 @@
...
+  useEffect(() => {
+    const timer = setTimeout(() => setDebouncedValue(value), delay || 500);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [value, delay]);
+
+  return debouncedValue;
+}
```
brossetti1, line 15: i feel like we shuld move this to the ui library hooks -- seems pretty generic across apps

apps/core-app/src/pages/Proposals.tsx
brossetti1: im having a hard time deciphering here - is `event` using the `debounceSearchTerm`?

brossetti1 (changes requested):
left some small comments

brossetti1 (approved):
approving so comments dont get in the way of merging

apps/core-app/src/utils/debounceHook.ts
jordanlesich: Agreed

apps/core-app/src/pages/Proposals.tsx
skuhlmann: Yep

PR #755, opened 2022-08-30
upgrading radix libraries

brossetti1:
## GitHub Issue

N/A

## Changes

we should probably do some quick testing on this -- the updates are listed here - https://www.radix-ui.com/docs/primitives/overview/releases - some breaking changes we should reveiw

updating radix libraries - they are now on version 1.0.0

discovered this when our tooltip did not match the docs and the interfaces i was trying to access

## Packages Added

## Checks

- [ ] App runs locally


0xRowdy:
Have you had a chance to test everything out after upgrading?

We can always put a pin in this until after MCon if it has the potential to cause breaking changes.

brossetti1:
@0xRowdy not yet -- yeah i figured we can do it after mcon -- ill add dont merge

skuhlmann (approved):

Issue #770, opened 2022-08-31
button size off


brossetti1:
### Feature

core-app

### App Affected

- [X] [Admin](https://admin.daohaus.fun)
- [ ] [Summon](https://summon.daohaus.fun)
- [ ] [Hub](https://hub.daohaus.fun)
- [ ] Toolbox

### Expected behavior

button should fill container

### Actual behavior

button is overflowing content

### Steps to reproduce

visit proposal page - fitlers button is off

### Screenshots and context

![image](https://user-images.githubusercontent.com/5998100/187753263-78878762-aaa2-4773-a608-b1ef9394a759.png)

PR #771, opened 2022-08-31
removing min-width

brossetti1:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/770

## Changes

- [X] let widht of button adjust by the content inside it

Screen Shot 2022-08-31 at 12 34 14 PM

## Packages Added

## Checks

- [ ] App runs locally


brossetti1:
@0xRowdy wondering if this could have any side effects, in general, i think these min/max width on the default css can cause issues like this -- better to remove them.

0xRowdy:
> @0xRowdy wondering if this could have any side effects (see image in issue), in general, i think these min/max width on the default css can cause issues like this -- better to remove them.

Yeah, I agree.
I wasn't a fan of how this button handled width originally. Two options. fit-content or 100%. This requires someone to create a parent div to control the size if they wanted a button that was 30px wide. I set min-width based on what width used to be. I don't expect any side effects. 

Should also be removed it from the sm & lg variants. Around line 120

libs/ui/src/components/atoms/Button/Button.styles.ts
```diff
@@ -17,7 +17,6 @@ export const ButtonBase = styled.button`
   height: 4.8rem;
   justify-content: left;
   letter-spacing: 1.8px;
-  min-width: 8.5rem;
```
0xRowdy, line 20: Also remove on sm and lg classes. Around line 120

Issue #805, opened 2022-09-02
Add in context form buttons


skuhlmann:
- pass local values in url?

- manage shaman
    - remove from proposal list
  
- governance
- token settings
-  transfers from vault
- issues share/loot from members
- guildkick from member if default value is there
-

PR #808, opened 2022-09-03
Feature/more buttons

skuhlmann:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/805

## Changes
- new proposal page parses default values
- add members
- transfer tokens
- settings forms

Screenshot from 2022-09-03 12-47-16
Screenshot from 2022-09-03 12-47-11
Screenshot from 2022-09-03 12-47-07

## Packages Added

n/a
## Checks

- [x] App runs locally


brossetti1 (approved):
looks good, are the new proposal default values going to be configurable for the dao? 

like proposal settings or something?

skuhlmann:
> looks good, are the new proposal default values going to be configurable for the dao?
> 
> like proposal settings or something?

We'll need to figure ok uot how to do that. Those URLs in the new proposal modal are built off form legos and we dont currently indicate defaults in the lego. Something we can think about next week.

Issue #713, opened 2022-08-24
Create Cancel Proposal Button in Proposal Details


skuhlmann:
# Design
Figma: https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=5591%3A120888

<img width="739" alt="Screen Shot 2022-08-29 at 12 53 09 PM" src="https://user-images.githubusercontent.com/42706432/187253335-e11201d4-472b-4290-a00b-a3cafea7bcf6.png">

# Cancel Proposal button
- ONLY DO THIS ON THE DETAILS PAGE
- only show if in voting + user _can_ cancel (must be sponsor) otherwise this button should not appear 
- only can cancel if:
```
       msg.sender == prop.sponsor ||
                sharesToken.getPriorVotes(prop.sponsor, block.timestamp - 1) <
                sponsorThreshold ||
                isGovernor(msg.sender),
```
   - user is the sponsor OR
   - user is a governor shaman
   - sponsor.shares < dao.sponsorThreshold
      - will require another fetch

earth2travis:
@adriennetilley are we able to get the designs done so we can implement this feature this week?

adriennetilley:
Designs done, just a simple button that only appears if you are capable of cancelling.

PR #831, opened 2022-09-13
Feature/713 cancel prop

skuhlmann:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/713

## Changes

subgraph
- proposer and sponsor entities in proposal so we can get ad shares easily

libs
- permission helper functions
- dao-data updates to add to the proposal queries

apps - core:
- cancel proposal component added to proposal details
- shows if in the voting period
- disabled if the connected address is able to cancel
- fires a cancel proposal tx

Screenshot from 2022-09-13 13-10-05

## Packages Added

n/a

## Checks

- [x] App runs locally


apps/core-app/src/pages/Members.tsx
```diff
@@ -197,13 +197,15 @@ export function Members() {
       actions={[
         <StyledButtonLink
           href={`/molochv3/${daochain}/${daoid}/new-proposal?formLego=ISSUE`}
+          key="1"
```
jordanlesich, line 200: I could be wrong about this, but if we use '1' or '2' as keys then react might confuse them with other lists with keys that have the same name, which could cause perf and state problems. 

Read here: 
https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/
https://reactjs.org/docs/lists-and-keys.html#keys

jordanlesich (approved):
Approved with changed to key pattern.

apps/core-app/src/pages/Members.tsx
skuhlmann: this one isn't inside a map - but i'll switch away from '1', '2' here.

skuhlmann: actually not going to put those in an array - just wrap in a fragment and send a single node through. feels better.

PR #832, opened 2022-09-13
Process next prop

jordanlesich:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/822

## Changes

Screenshot from 2022-09-13 14-59-17

Also fixed proposal card actions width

## Packages Added

NA

## Checks

- [x] App runs locally


apps/core-app/src/pages/ProposalDetails.tsx
```diff
@@ -51,23 +51,25 @@ export function ProposalDetails() {
   >();
   const [proposalLoading, setProposalLoading] = useState<boolean>(false);
 
+  const fetchProposal = useCallback(() => {
```
skuhlmann, line 54: We are going to have conflicts as i did something similar for my pr to refresh after the cancel proposal tx.

I need to pass a function down to use in the lifecycle. If we can pass this callback down that could replace what I did.

skuhlmann (approved):

apps/core-app/src/pages/ProposalDetails.tsx
skuhlmann: i fixed this

PR #845, opened 2022-09-16
Break up common utilities

brossetti1:
## GitHub Issue

NA - will add later

## Changes

split ethers dependencies up from common utiltiies

the dependency graph is really useful to show where these exist -- nx graph

the UI for example -- looks like its still dependent on `dao-data` -- not sure if this is in multiple places right now, but one thing we can do to reduce things is find those spots and make the `daodata` dependency values an argument (or some other strategy) so it unlinks the dependency from the UI --- making the build way smaller and the dependency's way less dependent on what the `daodata` relies on.

Screen Shot 2022-09-19 at 3 33 38 PM
Screen Shot 2022-09-19 at 3 33 44 PM
Screen Shot 2022-09-19 at 3 33 38 PM

core-app is building 1800kb file after I have included an optimization plugin for vite --- the other thing we can do use react-router to do dynamic import components which would only load these pieces of the app dynamically when the routes are hit -- i believe this reduces the package size a lot.

https://v5.reactrouter.com/web/guides/code-splitting

so in the core-app/src/Routes.tsx file you would do the following with the example in the react-router docs:

Screen Shot 2022-09-19 at 3 27 36 PM

```
const Routes = () => {
  return (
    <>
      <Banner />
      <RoutesDom>
        <Route path="/" element={<Home />} />
        <Route path="molochv3/:daochain/:daoid" element={<DaoContainer />}>
          <Route index element={<LoadableDashboard />} /> <---dynamic import on this route
          <Route index element={<DaoOverview />} />
          ...other routes
        </Route>
      </RoutesDom>
    </>
  );
};
```

## Packages Added

N/A

## Checks

- [ ] App runs locally


0xRowdy:
What is the current status and thoughts moving forward on this? Should I review and try and get merged or still WIP?

brossetti1:
@0xRowdy just seeing this - current state is needing a rebase and updating imports plus changing any adjusted functionality that has been added since the last commit here

I am going to break this up into 3 prs probably to help with the transition of breaking this up. I pobably wont get to it until this weekend or next week

earth2travis:
@brossetti1 are you going to be able to break these PRs up this week?

brossetti1:
@earth2travis yeah planning on it! finally getting some time this week to work on stuff

earth2travis:
Issues addressed with new monorepo

Issue #672, opened 2022-08-15
Build Rage Quit Feature and Form


earth2travis:
To be broken out of this design card:

https://github.com/HausDAO/daohaus-monorepo/issues/206

Figma:

https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6808%3A197739 

Ragequit workflow will be a form accessed from the Member Profile page. 

ParMd: "Step 1. Select voting and/or non-voting tokens"
ParXs: "The amount of voting and non-voting tokens selected in step 1 determines the amount of tokens available to ragequit in step 2."

Field - "Voting Tokens (TOKEN NAME)" 
- name of voting token in parenthesis 
- include a Max Amount indicator 
- numbers only, can include decimals
- info icon wording - ‘These tokens provide voting privilege and membership in the DAO’

Field - "Non-Voting Tokens (TOKEN NAME)" 
- name of non-voting token in parenthesis 
- include a Max Amount indicator 
- numbers only, can include decimals
- info icon wording: ‘These tokens provide non-voting, economic membership in the DAO’

horizontal rule

ParMd: "Step 2. Select tokens to ragequit"

Table:
Amounts in the table will update after values in the voting/non-voting token inputs are adjusted - after clicking out of those inputs, trigger refresh of table values

Table Headers:
- checkbox, checked by default
- Token
- Amount
- USD Value (tooltip - needs text about how value is estimate)

Table rows:
- checkboxes, checked by default
- token component - includes token symbol, name, copy icon, and link
- amount to 2 decimal spaces ?
- dollar symbol and amount to 2 decimal spaces ?

Vertical scrollbar to the right of table columns
- on mobile, table may need a horizontal scrollbar depending on breakpoint, but vertical scroll can be removed (uses same scroll as the page).

Load More button
- centered. stays centered on mobile; does not scroll sideways with tokens 
- primary outline button
- can maximize number of tokens that show before loading more - 20 or 30 would both be fine

horizontal rule

Checkbox - "Ragequit to different address (optional)"
- unchecked by default

Field - "Address to send funds"
- address field
- field shows conditionally when the checkbox above it is checked; when unchecked, it hides field. If form is submitted with a value in the field but the field hidden because the checkbox has been unchecked, clear the value and do not send funds to it.
- underneath the field, return ceramic profile avatar and name, or if that's unavailable, ENS info. If address belongs to a member of the DAO, display a "Member" tag

Submit button 
[...]

PR #865, opened 2022-09-19
Feature/672 rage quit

skuhlmann:
## GitHub Issue

fixes https://github.com/HausDAO/daohaus-monorepo/issues/672

## Changes

- ragequit page and route
- ragequit form and tx legos
- custom rq token input (shares or loot)
- custom rg token list select input

## Packages Added

n/a

## Checks

- [x] App runs locally


apps/core-app/src/components/customFields/RagequitToken.tsx
```diff
@@ -0,0 +1,55 @@
...
+    if (!dao || !connectedMembership) return null;
+    return {
+      label:
+        id === 'sharesToBurn'
+          ? `Voting Tokens (${dao.shareTokenSymbol})`
+          : `Non-Voting Tokens (${dao.lootTokenSymbol})`,
+      maxAmount:
+        id === 'sharesToBurn'
+          ? connectedMembership.shares
+          : connectedMembership.loot,
```
brossetti1, line 24: question about loot and shares -- can a member only have 1 or the other, or can they have both?

if both, just curious-- would we want to present that here?

skuhlmann: They can have both - this input is used in both instances - so we have it in the form twice - once for shares and once for loot.

```ts
       fields: [
          {
            id: 'sharesToBurn',
            // @ts-expect-error: cannot resolve type within formSegment using custom fields sent from the core-app
            type: 'ragequitToken',
          },
          // @ts-expect-error: cannot resolve type within formSegment using custom fields sent from the core-app
          { id: 'lootToBurn', type: 'ragequitToken' },
        ],
        ```

File: apps/core-app/src/components/customFields/RagequitTokenList.tsx [hunk 2]
```diff
@@ -0,0 +1,212 @@
...
+  };
+
+  const tokenTable = useMemo((): TokenTable | null => {
+    if (!dao || !networkData || !connectedMembership) return null;
+    return dao?.tokenBalances
+      .filter((token) => Number(token.balance) > 0)
+      .reduce(
+        (acc: TokenTable, token: TokenBalance) => {
+          acc.tokenCheckboxes = [
+            ...acc.tokenCheckboxes,
```
brossetti1, line 95: im going to have to a get a breakdown of how you use reduce -- i usually pick other methods but this is really nice

apps/core-app/src/legos/form.ts
```diff
@@ -543,4 +543,50 @@ export const COMMON_FORMS: Record<string, CustomFormLego> = {
...
+        fields: [
+          {
+            id: 'sharesToBurn',
+            // @ts-expect-error: cannot resolve type within formSegment using custom fields sent from the core-app
+            type: 'ragequitToken',
+          },
+          // @ts-expect-error: cannot resolve type within formSegment using custom fields sent from the core-app
+          { id: 'lootToBurn', type: 'ragequitToken' },
+        ],
+      },
```
brossetti1, line 564: so i beleive if you add these fields to https://github.com/HausDAO/daohaus-monorepo/blob/develop/libs/form-builder/src/components/CoreFieldLookup.ts#L19 you do not need the `@ts-expect-error`

brossetti1 (approved):
added some comments -- approving to not block this but id review the types and ts commetns

libs/common-utilities/src/constants/contracts.ts
```diff
@@ -12,6 +12,9 @@ export type ContractKey =
 export type AddressKeyChain = { [key in ValidNetwork]?: string };
 export type ContractAddressList = Record<ContractKey, AddressKeyChain>;
 
+export const NETWORK_TOKEN_ETH_ADDRESS =
+  '0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE';
+
```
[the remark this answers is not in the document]
skuhlmann: > what does this value do?
> 
> I noticed its used as a fallback for alot of different values

This is used in Baal for the network token address so it's needed in rage quit.

apps/core-app/src/legos/form.ts
skuhlmann: right but these are all fields in the form builder itself - not in the core app.

Issue #881, opened 2022-09-21
disconnect daodata from UI components


brossetti1:
## Remove DaoData from UI components

https://github.com/HausDAO/daohaus-monorepo/blob/master/libs/ui/src/components/organisms/MemberCard/MemberCard.tsx#L9

The UI contains 1 import from `@daohaus/daodata` - removing this unlinks ethers from the UI via ethers-utilities and removes any dependencies coming from `daodata` -- making our UI package alot smaller.

earth2travis:
Do you still plan on finishing this @brossetti1?

PR #882, opened 2022-09-21
removing dao-data from ui

brossetti1:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/881

## Changes

one thing we can think about doing is creating a `types` folder under libs and nest different types in that folder - we can export that entire folder as one package AND split those nested types up into their own packages, that way people can choose which package to include, similar to how `ethers` does this.

https://github.com/ethereumjs/ethereumjs-monorepo#packages

you can see their dependency tree here

https://github.com/ethereumjs/ethereumjs-monorepo#package-dependency-relationship

## Packages Added

## Checks

- [ ] App runs locally

### DIFF
--- a/libs/ui/src/components/organisms/MemberCard/MemberCard.tsx
+++ b/libs/ui/src/components/organisms/MemberCard/MemberCard.tsx
@@ -6,7 +6,7 @@ import {
   truncateAddress,
   ValidNetwork,
 } from '@daohaus/common-utilities';
-import { AccountProfile } from '@daohaus/dao-data';
+import { AccountProfile } from '@daohaus/common-utilities';
 
 import { ParMd } from '../../atoms';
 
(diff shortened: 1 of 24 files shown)


jordanlesich:
What's the blocker here? Is this something I can finish off?

brossetti1:
@jordanlesich sorry not sure why i didnt see this message, getting time this week to finish this

libs/ui/src/components/organisms/MemberCard/MemberCard.tsx
```diff
@@ -6,7 +6,7 @@ import {
   truncateAddress,
   ValidNetwork,
 } from '@daohaus/common-utilities';
-import { AccountProfile } from '@daohaus/dao-data';
+import { AccountProfile } from '@daohaus/common-utilities';
```
brossetti1, line 9: this is the main change of this pr -- moving the type into common utilities removes dao-data from the UI lib

skuhlmann (approved):

Issue #160, opened 2022-05-16
 Build Wallet Connect Proposal Form


skuhlmann:
# Overview

WalletConnect proposal form will allow you to connect a baal gnosis safe to a dapp and make a proposal based on that interaction.

## Use cases

A user can create a proposal using wallet connect
- open a wallet connect proposal form
- visit a dapp that supports wallet connect, initiate a connection and copy the connection code
- paste that code into the dh proposal form
- initiate a transaction in the dapp and the DH form will catch it and react showing the transaction details
- user can then submit the proposal and it'll fill in that action data to be executed upon a passed proposal

A user can see a proposal card/detail of a wallet connect proposal
- the proposal detail now will just look like any other arbitrary contract call

All other action will be the same as other proposals
- A member can sponsor the proposal, vote and execute

## Design

Figma: https://www.figma.com/file/9lCwprCNaoAmUDmiMQpFEA/DAOhaus-v3?node-id=6346%3A179339

## General Considerations

-demo video of v2 implementation
https://www.youtube.com/watch?v=KFjuf5Sz7Sk
  ![Screenshot from 2022-05-16 14-11-40](https://user-images.githubusercontent.com/6923345/168674494-4f5f915d-8a8b-4814-ba55-dfcdec08ff46.png)

- We now need to provide an estimated gas cost for executing the proposal action with each proposal. Our plan is to use the gnosis api behind the scenes to get that estimation.
  - do we want to display that to the user?
  - let them adjust that?
  - let them select 0, which, I think will override the estimation and let the action execute with no gas limit

### Form details/fields

- Name
- Description
- Link/content url
- Proposal expiration time
- Wallet Connect interaction/display

### Form Considerations

- The proposal creation process is the main interaction, once submitted the proposal will look like any ACE prop.

## Prototype Items to Capture

earth2travis:
@adriennetilley any design left on this one?

earth2travis:
Get the form working in core ui and then decide if it should be a part of TX Builder or a separate app

santteegt:
This is how it currently looks:

* Stage 1: specifying WalletConnect link to a dApp
![image](https://user-images.githubusercontent.com/926341/191056761-967a7afd-dbae-41d1-975f-718f64898f90.png)

* Stage 2: Waiting for Tx to be triggered on the connected dApp
[...]

PR #899, opened 2022-09-23
Add WalletConnect proposal form

santteegt:
## GitHub Issue

#160 

## Changes

Integrates WalletConnect to a proposal form to connect a DAO Safe to any dApp and trigger transactions.

## Packages Added

- buffer@^6.0.3

## Checks

- [X] App runs locally


libs/tx-builder-feature/src/utils/multicall.ts
```diff
@@ -139,8 +139,15 @@ export const handleMulticallArg = async ({
       if (data) {
         return {
           to: processedContract.address,
-          data,
-          value: Number(processValue),
+          data: isSearchArg(data)
```
jordanlesich, line 142: This seems odd to me. So if it's not a search arg, then it definitely has a .value & is definitely a string? 

If so, then ignore this comment.

jordanlesich (commented):
Looking great! Just a few small-ish changes and we're good to go.

skuhlmann (approved):

Issue #862, opened 2022-09-19
Home nav item stays selected even when on different page


7helles:
### Feature

Admin top nav

### App

- [X] [Admin](https://admin.daohaus.fun)
- [ ] [Summon](https://summon.daohaus.fun)
- [ ] [Hub](https://hub.daohaus.fun)
- [ ] [Toolbox](https://toolbox.daohaus.fun/)

### Expected Behavior

When navigating from one page to another in Admin, only one page (the current location of the user) should be active/selected.

### Actual Behavior

The navigation bar at the top of Admin looks like "Home" is active even when the user has navigated to another page.

### Steps to Reproduce

1. Go to a DAO Home page, like https://admin.daohaus.fun/#/molochv3/0x64/0x68ee75440c79e6395880ea8009c9307432e5630b
2. Navigate to another page, such as Proposals
3. See how the "Home" nav item is still selected

jordanlesich:
This is related to a bad useLocation pattern in DH Connect.

Issue #844, opened 2022-09-16
Hub App Dao Card Button


0xRowdy:
### Feature

Dao List Cards

### App

- [ ] [Admin](https://admin.daohaus.fun)
- [ ] [Summon](https://summon.daohaus.fun)
- [X] [Hub](https://hub.daohaus.fun)
- [ ] [Toolbox](https://toolbox.daohaus.fun/)

### Expected Behavior

You should only be ale to click on the button.

### Actual Behavior

The clickable area is the whole width of the card. See screenshot

![Screen Shot 2022-09-14 at 1 03 31 PM](https://user-images.githubusercontent.com/22626267/190713674-68caf68b-2cdf-46e5-9e47-841b67e37ca9.png)

### Steps to Reproduce

1. Go to Hub
2. Authenticate with a wallet
* Must have DAOs listed
4. Hover over DAO Card "Go" Button. Hover and click to the right side of the button.

### Browser

Firefox, Chrome, Brave, Safari

jordanlesich:
I'd say wait for this. I'm thinking we're going to build Hub into admin

jordanlesich:
Made the button fullwidth, as per another card. Fixes this issue.

Issue #661, opened 2022-08-15
Hub: DAO Card Mobile View Improvements


jonathanprozzi:
### Feature

Exploring the Hub app on a phone/tablet

### Expected outcome

For the UI to be fully usable on phone or tablet

### Actual outcome

Only slight changes need to be made for mobile improvements:

- DAO Cards
 - Make buttons full width
 - Alignment changes:
   -  Filter full with to match the button
   - Sort by left aligned
   - DAO names 2 lines after a max width (before truncating)

### Browser

Brave

### Screenshots and context

<img width="331" alt="image" src="https://user-images.githubusercontent.com/9438776/184636793-37aedc24-9819-4d01-b109-a95e6dbaf94f.png">

jonathanprozzi:
@earth2travis going to self-assign this as well

Issue #596, opened 2022-08-02
Update Icon on Hub Card


earth2travis:
From Helle

i noticed a very small thing - on hub, the card view/list view button probably should change its icon depending on whether it says "card" or "list", so with "list" maybe the icon could be the "menu-line" icon

![Image](https://user-images.githubusercontent.com/522182/179835649-a265e075-ec5e-42dd-81f8-957c8ae003e3.png)

![Image](https://user-images.githubusercontent.com/522182/179835695-712b3c38-2c28-4e3c-865c-bdeb04ab197e.png)

i don't have any DAOs showing so i'm not sure how it currently works, but i'm assuming it shows "list view" when cards are being displayed, and vice versa for when the list is displayed, so users know what they can toggle to

REACT/REMIX ICON:

`RiMenuFill`

earth2travis:
@7helles can you please provide a link to the icon in Figma?

jordanlesich:
Let's halt improvements to Hub

PR #928, opened 2022-10-03
Hub integration

jordanlesich:
## GitHub Issue

#538 
#596 
#661 
#410 
#844 
#538 
#536 
#862

## Changes

- Fixes most of the above issues or ensures that the issues are no longer relevant with current rebuild. 
- Transplants (and partially rebuilds) code from hub
- Refactors some components of hub to reusable components
- Fixes navigation issues (link 

## Profile Page
I removed Hub's profile page for the following reasons

- Hub's profile page was only display DAOs for a single network. 
- The DAO display functionality is the exact same as Hub's 
- If we want a publicly viewable page, couldn't we just add an address to hub url? 
- To my knowledge, Lens protocol is still in whitelist phase and there's no real way to onboard for their mainnet. 

Let's discuss on Monday. 

## Packages Added

## Checks

- [x] App runs locally


apps/core-app/src/components/DaoTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+import { ProfileAvatar, Tag } from '@daohaus/ui';
+import {
+  charLimit,
+  getNetworkName,
+  readableNumbers,
+  truncateAddress,
+} from '@daohaus/common-utilities';
+
+// REVIEW NOTES
+// Can this be refactored to use DaoHaus Table?
```
skuhlmann, line 16: It probably can be.

You have the prep the columns and table data up above and the pass a few other things:
- array of sortable column ids
- load more method
- sort method

We could track this in a new refactor ticket

apps/core-app/src/Routes.tsx
```diff
@@ -15,13 +22,59 @@ import ProposalDetails from './pages/ProposalDetails';
...
+  const navigate = useNavigate();
+  const { isConnected, address } = useHausConnect();
+  const { profile } = useParams();
+
+  useLayoutEffect(() => {
+    if (isConnected && address && profile) {
+      return;
+    }
+    if (isConnected && !profile) {
+      navigate(`/${address}`);
```
skuhlmann, line 43: Wondering if navigation on connection is what we'd want here.

If we just kept them at the base route, but didn't show the hub data stuff would it be better?
Would enable a user to visit another addresses profile, right?

skuhlmann (approved):

Issue #969, opened 2022-10-12
include package dependencies in build


brossetti1:
# Problem

our builds do not currently include the package dependencies for a few libraries.

@keating fixed this on his own vite plugin and presented a working example:

https://github.com/HausDAO/daohaus-monorepo/compare/develop...feature/generate-deps - including extension and cleaning up monorepo build
https://github.com/nxext/nx-extensions/pull/801/files - nx extension update

We should bring these changes into the monorepo

# Solution

- [ ] create the executor for the custom packageJson update
- [ ] include changes for vite builds
- [ ] cleanup packageJsons as seen in @keatings `generate-deps` pr

earth2travis:
Addressed with new monorepo

PR #970, opened 2022-10-12
Keating build upgrade

brossetti1:
## GitHub Issue

https://github.com/HausDAO/daohaus-monorepo/issues/969

## Changes

- [X] create the executor for the custom packageJson update
- [X] include changes for vite builds
- [X] cleanup packageJsons as seen in @keatings `generate-deps` pr

The executor is stored in `tools/executors` - here is a better idea of where the executor plugs in:

Screen Shot 2022-10-12 at 12 17 28 PM
Screen Shot 2022-10-12 at 12 18 17 PM
Screen Shot 2022-10-12 at 12 20 04 PM

## Packages Added

None

## Checks

- [ ] App runs locally


brossetti1:
@alexkeating can you check this work

basically pulled alot of it from your branches from a while back:

https://github.com/HausDAO/daohaus-monorepo/compare/develop...feature/generate-deps
https://github.com/nxext/nx-extensions/pull/801/files

libs/common-utilities/project.json
```diff
@@ -16,6 +16,14 @@
...
       }
     },
+    "vite-build": {
+      "dependsOn": [
+        {
+          "target": "build",
+          "projects": "self"
+        }
+      ]
+    },
```
brossetti1, line 26: not sure if this is needed

libs/dao-data/project.json
```diff
@@ -22,6 +22,14 @@
...
       }
     },
+    "vite-build": {
+      "dependsOn": [
+        {
+          "target": "build",
+          "projects": "self"
+        }
+      ]
+    },
```
brossetti1, line 32: not sure if this is needed - dao-data does not use vite

libs/daohaus-connect-feature/project.json
```diff
@@ -5,14 +5,17 @@
...
       "options": {
         "outputPath": "dist/libs/daohaus-connect-feature",
         "frameworkConfigFile": "@nxext/react/plugins/vite",
         "configFile": "libs/daohaus-connect-feature/vite.config.ts",
         "entryFile": "src/index.ts",
-        "generatePackageJson": true
+        "updateBuildableProjectDepsInPackageJson": true,
+        "buildableProjectDepsInPackageJsonType": "dependencies",
+        "packageJson": "libs/daohaus-connect-feature/package.json",
+        "project": "libs/daohaus-connect-feature"
```
brossetti1, line 18: here is the change where we add our exectutor from `tools/executor` --- is this naming ok? 

let me know if you think of a better one

libs/dao-data/project.json
skuhlmann: yeah that doesn't seem necessary

skuhlmann (commented):
looking good - we can test with an external app soon.

libs/dao-data/project.json
brossetti1: hmm weird, looks like we do need this?

got to be something we can do about it -- ill look into it

Screen Shot 2022-10-12 at 5 02 22 PM

alexkeating: That error basically means that command needs to exist for libraries that it depends on. That's why I added it here

libs/daohaus-connect-feature/project.json
alexkeating: Might be better to name it @daohaus/nx-plugin as I think the package could contain both executors and generators

alexkeating (commented):
Looks good to me

skuhlmann (approved):

PR #1013, opened 2022-10-19
Use tx metadata to parse abi encoded params

### LINKED ISSUE #942: Further Decode Action Data on GovernanceConfig Proposals
# Details

When decoding action data for governance setting update proposals we need to do a 2nd round of decoding so we can display the args in the governance config.

These are encoded in when creating the proposal.

Example:
https://admin.daohaus.fun/#/molochV3/0x5/0x5e0158adb4d5ccf34736aa5e3d25b23ce41c131d/proposals/38

Image

Contract reference:
https://github.com/HausDAO/Baal/blob/feat/baalZodiac/contracts/Baal.sol#L844 

```
    function setGovernanceConfig(bytes memory _governanceConfig)
        external
        baalOrGovernorOnly
    {
        (
            uint32 voting,
            uint32 grace,
            uint256 newOffering,
            uint256 quorum,
            uint256 sponsor,
            uint256 minRetention
        ) = abi.decode(
                _governanceConfig,
                (uint32, uint32, uint256, uint256, uint256, uint256)
            );
```

santteegt:
## GitHub Issue

Closes #942 

## Changes

Use tx lego metadata to parse ABI encoded params. For example, Baal governance config params are sent as ABI-encoded bytes

## Packages Added

None

## Checks

- [X] App runs locally

### DIFF
--- a/libs/tx-builder-feature/src/utils/decoding.ts
+++ b/libs/tx-builder-feature/src/utils/decoding.ts
@@ -3,10 +3,14 @@ import {
   ArgType,
   CONTRACTS,
   ENCODED_0X0_DATA,
+  MulticallAction,
+  StringSearch,
+  ValidArgType,
   ValidNetwork,
 } from '@daohaus/common-utilities';
 import { LOCAL_ABI } from '@daohaus/abi-utilities';
 import { createContract, fetchABI } from './abi';
+import { isSearchArg } from './args';
 
 const OPERATION_TYPE = 2;
 const ADDRESS = 40;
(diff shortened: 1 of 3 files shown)


libs/tx-builder-feature/src/utils/decoding.ts
```diff
@@ -110,12 +114,36 @@ const buildEthTransferAction = (action: EncodedAction): DecodedAction => ({
   params: [],
 });
 
+const decodeParam = ({
+  argMeta,
+  value,
+} : {
+  argMeta?: ValidArgType,
+  value: any,
```
skuhlmann, line 122: lets put a ts lint ignore here

skuhlmann (changes requested):

skuhlmann (approved):

